PR #9, opened 2022-03-14
UefiPayloadPkg/Logo/Logo.c: Fallback to builtin logo

mkopec:
Signed-off-by: Michał Kopeć <michal.kopec@3mdeb.com>


mkopec:
Made `UefiPayloadPkg/Logo/Logo.bmp` a symlink to `MdeModulePkg/Logo/Logo.bmp` - turns out git handles symlinks correctly. @miczyg1 please let me know if this is acceptable

UefiPayloadPkg/UefiPayloadPkg.dsc
```diff
@@ -557,7 +557,7 @@
 
   UefiCpuPkg/CpuDxe/CpuDxe.inf
   MdeModulePkg/Universal/BdsDxe/BdsDxe.inf
-  UefiPayloadPkg/Logo/Logo.inf
+  MdeModulePkg/Logo/Logo.inf
```
miczyg1: Isn't this wrong? It won't compile your custom Logo driver now @mkopec 
Here and below applies the same comment

miczyg1 (changes requested):

UefiPayloadPkg/UefiPayloadPkg.dsc
mkopec: This .inf only describes the bmp file, i thought that if include MdeModulePkg's version here, the bitmap from there would be used in UefiPayloadPkg. Turns out it doesn't make a difference.

Reverted to `UefiPayloadPkg/Logo/Logo.inf`.

miczyg1: In that case both versions should work although it would have to be tested

PR #16, opened 2022-05-17
Add TPM Physical Presence Interface

miczyg1:


miczyg1:
@mkopec would you mind testing it on any TGL laptop? I wonder if it doesn't brake anything.

mkopec:
Seems to work okay. I can see PPI options in UEFIPayload setup menu and a folder `/sys/class/tpm/tpm0/ppi` has appeared

PR #17, opened 2022-05-24
Display popup if firmware recovery was used

SergiiDmytruk:
Had to make some of `MdeModulePkg/Universal/DisplayEngineDxe/` functions public in `CustomizedDisplayLib`. Another addition is parsing of vboot workbuf in `UefiPayloadPkg/Library/CbParseLib/`.

Popup displays number of seconds left before boot continues, which has an annoying effect of cursor movement because at least in `minicom` cursor is not hidden despite use of API to hide it. It probably works differently locally.


UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
```diff
@@ -735,3 +737,219 @@ ParseTPMPPIInfo (
...
 }
+
+STATIC
+CONST CHAR8 *
+GetRecoveryReasonString(
+  IN UINT8 code
+  )
+{
+  switch ((enum vb2_nv_recovery)code) {
+  /* 0x00 */ case VB2_RECOVERY_NOT_REQUESTED:
```
miczyg1: Maybe the comment with hex codes should be at the end of line? Now it makes the switch hard to read

UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
```diff
@@ -735,3 +737,219 @@ ParseTPMPPIInfo (
...
+RETURN_STATUS
+EFIAPI
+ParseVBootWorkbuf (
+  OUT UINT8        *RecoveryCode,
+  OUT CONST CHAR8 **RecoveryReason
+  )
+{
+  struct cb_cbmem_entry *CbmemEntry;
+  struct cb_vboot_workbuf_v2 *Workbuf;
+
```
miczyg1, line 933: Any NULL checks for input pointers?

miczyg1 (changes requested):

miczyg1 (approved):

PR #21, opened 2022-06-28
UefiPayloadPkg: Add PCD for default SecureBoot state

mkopec:
Add PCD to make default Secure Boot enablement controllable. Default to TRUE to maintain compatibility.

Signed-off-by: Michał Kopeć <michal.kopec@3mdeb.com>


mkopec:
Tested:

Build with the usual build options and see that Secure Boot is enabled by default.

Build with `-D SECURE_BOOT_DEFAULT_ENABLE=FALSE` and see that Secure Boot is disabled by default, and can be enabled by the user manually.

miczyg1 (approved):

PR #24, opened 2022-09-13
add debug prints that made optiplex booting

pietrushnic:
No idea, why but those debug prints make Dell OptiPlex 9010 booting, so there have to be some timing during initialization. There are other asserts I get while trying debug build (e.g. UiApp assert reported on Matrix). For now I'm using this code to move forward, but we have to find root cause of the problem. Question is also if things behave the same way on release build. What I tried was `config.dell_optiplex_9010.uefi.debug` only.

Signed-off-by: Piotr Król <piotr.krol@3mdeb.com>


miczyg1:
Apparently, there is some timing issue when EDK2 is issuing SMMSTORE SMIs to write to SPI flash. I also had to add like delays between writes and erases in UefiPayloadPkg SMMSTORE code

miczyg1:
I don't mind having a few of those until we find a root cause of it.

miczyg1 (approved):

PR #41, opened 2023-01-10
Updated smmstore

miczyg1:


miczyg1:
Fixes the random problems with SMMStore on Novacusto mADL laptops. Needs testing for all other platforms using UEFIPayload...

macpijan:
> Needs testing for all other platforms using UEFIPayload...

So far tested on ADL laptops only?

miczyg1:
@macpijan yup, I want to test Protectli VP46xx and MSI today as well.

miczyg1:
Checked MSI DDR4 and VP4650 and it works. I think we may fairly assume that the driver update is more robust and some bugs have been fixed.

mkopec (approved):

PR #43, opened 2023-01-11
Memory map fixes

miczyg1:
Fixes issues with certain dGPUs DMA errors. The UEFI memory map is incorrectly constructed. Pei driver reports all coreboot reserved resources (including MMIO) as reserved memory instead of MMIO. So MMIO and reserved memory are mixed together. The PR fixes that by checking if the resource is below TOLUD for reserved memory, if above it means MMIO. VT-d driver checks for the UEFI memory map and sets correct DMA protections, but only for memory. Before the patch it incorrectly set the protections for whole address space below 4G, now it sets the protection correctly. The TOLUD check is also required because RMRR is placed below TOLUD and it has to be reported as reserved memory, otherwise, OS complained on incorrect mapping before the fix and things could break.

### DIFF
--- a/UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
+++ b/UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
@@ -390,7 +390,6 @@ ParseMemoryInfo (
   MEMROY_MAP_ENTRY         MemoryMap;
   UINT32                   Tolud;
 
-
   Tolud = PciRead32(PCI_LIB_ADDRESS(0,0,0,0xbc)) & 0xFFF00000;
 
   //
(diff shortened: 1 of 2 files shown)


UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
```diff
@@ -388,6 +388,10 @@ ParseMemoryInfo (
   struct cb_memory_range   *Range;
   UINTN                    Index;
   MEMROY_MAP_ENTRY         MemoryMap;
+  UINT32                   Tolud;
+
+
```
mkopec:
```suggestion
```

Extra empty line?

mkopec (commented):
Looks good, just one nitpick

mkopec (approved):

PR #55, opened 2023-03-21
UefiPayloadPkg/Library/PlatformBootManagerLib: add SOL strings

miczyg1:


mkopec:
I guess if we want to use this with a platform without an EC, we should guard the EC FW checks with an additional PCD

miczyg1:
> I guess if we want to use this with a platform without an EC, we should guard the EC FW checks with an additional PCD

No need for that for now. coreboot does not produce SMBIOS type 11 for any platform and doesn't seem to ever be done. We can also use a custom SMBIOS type number to convey such info

mkopec:
Ack, thanks.

mkopec (approved):

PR #58, opened 2023-06-13
Nvc boot optimization

### LINKED ISSUE #454: SD card is not recognized as bootable
**Dasharo version**
v1.4.0
https://docs.dasharo.com/variants/novacustom_nv4x_tgl/releases/#v140-2023-02-24

**Affected component(s) or functionality**
Booting into kernel stored on SD card

**Brief summary**
I'm unable to boot into Linux from SD card, it seems the card is not recognized. No corresponding menu option is rendered in any place where bootable USB sticks normally show up (alongside the always-present iPXE Network Boot and UEFI Shell):
- Boot Manager Menu (F7)
- Setup (F2) -> One Time Boot
- Setup (F2) -> Boot Maintenance Manager -> Boot Next Value
- Setup (F2) -> Boot Maintenance Manager -> Boot Options -> Change Boot Order

**How reproducible**
Always.

**Steps to reproduce the behavior**
1. Plug bootable SD card into the SD card slot (example partitioning scheme given below)
2. Power on device
3. Try one of the options listed in the *Brief summary* section
4. Menu entry for the SD card is missing

**Expected behavior**
A menu option for the SD card should be shown, and if chosen, the device should boot into the kernel stored on the card.

**Actual behavior**
No menu option for the SD card is generated, thus it is not possible to boot from SD card.

**Additional context**

Partitioning scheme:
`# parted /dev/mmcblk0 -- print` (executed from `archlinux-2023.06.01-x86_64.iso`)
```text
Model: SD SH64G (sd/mmc)
Disk /dev/mmcblk0: 63.9GB
Sector size (logical/physical): 512B/512B
Partition Table: gpt
Disk Flags: 

Number  Start   End    Size    File system  Name    Flags
 1      1049kB  512MB  511MB   fat32        ESP     boot, esp
 2      512MB   528MB  16.8MB               header
```
_Dasharo firmware logs_
SD card plugged in before boot, booting into [DTS 1.2.0](https://3mdeb.com/open-source-firmware/DTS/v1.2.0/dts-base-image-v1.2.0.wic.gz) using an USB stick plugged into the USB-A port on the right side of the device.
`cbmem -1`:
[cbmem_1_with_sdcard.log](https://github.com/Dasharo/dasharo-issues/files/11655796/cbmem_1_with_sdcard.log)

No SD card plugged in, booting into [DTS 1.2.0](https://3mdeb.com/open-source-firmware/DTS/v1.2.0/dts-base-image-v1.2.0.wic.gz) using an USB stick plugged into the USB-A port on the right side of the device.
`cbmem -1`:
[cbmem_1_without_sdcard.log](https://github.com/Dasharo/dasharo-issues/files/11655804/cbmem_1_without_sdcard.log)

_Additional notes_
- Flashing exactly the same image (including the partitioning scheme) to an USB stick and booting that works fine
- The SD card is detected and can be read and written to fine after boot into `nixos-minimal-23.05.553.e7603eba51f-x86_64-linux.iso` or `archlinux-2023.06.01-x86_64.iso` from USB stick
- The SD card is not detected after boot into  [DTS 1.2.0](https://3mdeb.com/open-source-firmware/DTS/v1.2.0/dts-base-image-v1.2.0.wic.gz) via USB stick. `dmesg -w` logs nothing new when the card is unplugged and plugged in again
- I am running the official Dasharo v1.4.0 for NV40MZ TGL here, however, the IME is disabled, which is normally not the case. This is because I ran a custom modification of v1.4.0 before that I used to disable the IME (see Matrix chat), and the IME does not seem to be automatically enabled again by the official v1.4.0. I did not test if booting SD cards worked fine before I disabled the IME
- SD card is a SanDisk Max Endurance 64GB MicroSD XC card (with passive adaptor)

Happy to collect more information if necessary.

**Solutions you've tried**

An different SD card with a different partitioning scheme (MBR-based) is also not recognized (this is `archlinux-2023.06.01-x86_64.iso` flashed to the card, which works via a USB stick):
```text
Model: SD SH64G (sd/mmc)
Disk /dev/mmcblk0: 63.9GB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags: 

Number  Start  End    Size    Type     File system  Flags
 2      813MB  828MB  15.7MB  primary  fat16        esp
```

miczyg1:
The PR introduced the following changes:

1. Backports an option to control SD/MMC timeout via PCD at build time (integrated in coreboot). By setting a lower timeout on SD/MMC init, it fixes the problem with long boot time on NV4x. However it does not fix https://github.com/Dasharo/dasharo-issues/issues/454
2. Reverts the change  https://github.com/Dasharo/edk2/commit/ca98d24b188813d1a919a8c75417910630c6f9d2 which broke the Continue hotkey (the enter key was passed e.g. to GRUb bootloader and one was unable to control the menu as GRUB got the key and booted its first priority)
3. Refreshes the logo at ReadyToBoot event. It results in a clear screen if Continue hotkey is used, instead of leaving half of the progress bar and strings left on the screen.

### DIFF
--- a/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -1088,7 +1088,9 @@ PlatformBootManagerAfterConsole (
   Black.Blue = Black.Green = Black.Red = Black.Reserved = 0;
   White.Blue = White.Green = White.Red = White.Reserved = 0xFF;
 
+  gST->ConOut->EnableCursor (gST->ConOut, FALSE);
   gST->ConOut->ClearScreen (gST->ConOut);
+
   WarnIfRecoveryBoot ();
 
   BootLogoEnableLogo ();
--- a/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -638,7 +638,6 @@ PlatformBootManagerBeforeConsole (
   VOID
 )
 {
-  EFI_INPUT_KEY                  Enter;
   EFI_INPUT_KEY                  Esc;
   EFI_INPUT_KEY                  F12;
   EFI_BOOT_MANAGER_LOAD_OPTION   BootOption;
@@ -649,12 +648,6 @@ PlatformBootManagerBeforeConsole (
   BOOLEAN                        BootMenuEnable;
   UINTN                          VarSize;
 
-  //
-  // Register ENTER as CONTINUE key
-  //
-  Enter.ScanCode    = SCAN_NULL;
-  Enter.UnicodeChar = CHAR_CARRIAGE_RETURN;
-  EfiBootManagerRegisterContinueKeyOption (0, &Enter, NULL);
   //
   // Map ESC to Boot Manager Menu
   //
@@ -1084,6 +1077,7 @@ PlatformBootManagerAfterConsole (
   BOOLEAN                        BootMenuEnable;
   UINTN                          VarSize;
   EFI_EVENT                      Event;
+  EFI_INPUT_KEY                  Enter;
 
   Black.Blue = Black.Green = Black.Red = Black.Reserved = 0;
   White.Blue = White.Green = White.Red = White.Reserved = 0xFF;
@@ -1094,6 +1088,13 @@ PlatformBootManagerAfterConsole (
   WarnIfRecoveryBoot ();
 
   BootLogoEnableLogo ();
+
+  //
+  // Register ENTER as CONTINUE key
+  //
+  Enter.ScanCode    = SCAN_NULL;
+  Enter.UnicodeChar = CHAR_CARRIAGE_RETURN;
+  EfiBootManagerRegisterContinueKeyOption (0, &Enter, NULL);
 
   // FIXME: USB devices are not being detected unless we wait a bit.
   gBS->Stall (100 * 1000);
--- a/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -1040,6 +1040,26 @@ PrintSolStrings (
 }
 
 /**
+  Refresh the logo on ReadyToBoto event. It will clear the screen from strings
+  and progress bar when timeout is reached or continue key is pressed.
+
+  @param    Event          Event pointer.
+  @param    Context        Context pass to this function.
+**/
+VOID
+EFIAPI
+RefreshLogo (
+  IN EFI_EVENT    Event,
+  IN VOID         *Context
+  )
+{
+  gBS->CloseEvent (Event);
+  gBS->Stall (100 * 1000);
+  gST->ConOut->ClearScreen (gST->ConOut);
+  BootLogoEnableLogo ();
+}
+
+/**
   Do the platform specific action after the console is connected.
 
   Such as:
@@ -1063,6 +1083,7 @@ PlatformBootManagerAfterConsole (
   BOOLEAN                        NetBootEnabled;
   BOOLEAN                        BootMenuEnable;
   UINTN                          VarSize;
+  EFI_EVENT                      Event;
 
   Black.Blue = Black.Green = Black.Red = Black.Reserved = 0;
   White.Blue = White.Green = White.Red = White.Reserved = 0xFF;
@@ -1144,6 +1165,13 @@ PlatformBootManagerAfterConsole (
     Print (L"%-5s to enter Boot Manager Menu\n", BootMenuKey);
 
   Print (L"ENTER to boot directly\n");
+
+  EfiCreateEventReadyToBootEx (
+             TPL_CALLBACK,
+             RefreshLogo,
+             NULL,
+             &Event
+             );
 }
 
 /**
@@ -1174,12 +1202,6 @@ PlatformBootManagerWaitCallback (
     (Timeout - TimeoutRemain) * 100 / Timeout,
     0
     );
-
-  if (TimeoutRemain == 0) {
-    gBS->Stall (100 * 1000);
-    gST->ConOut->ClearScreen (gST->ConOut);
-    BootLogoEnableLogo ();
-  }
 }
 
 /**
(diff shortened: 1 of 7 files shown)


miczyg1:
The PR breaks the vboot pop-up. When enter is pressed to skip the window, it also skips the menu prompt. it seems like the key was not dequeued.

miczyg1:
> The PR breaks the vboot pop-up. When enter is pressed to skip the window, it also skips the menu prompt. it seems like the key was not dequeued.

Fixed by https://github.com/Dasharo/edk2/pull/58/commits/d435d5ec7a77a1357f20ac21df6206ecab285247

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1063,18 +1076,29 @@ PlatformBootManagerAfterConsole (
...
   WarnIfRecoveryBoot ();
 
   BootLogoEnableLogo ();
 
+  //
+  // Register ENTER as CONTINUE key
+  //
+  Enter.ScanCode    = SCAN_NULL;
+  Enter.UnicodeChar = CHAR_CARRIAGE_RETURN;
+  EfiBootManagerRegisterContinueKeyOption (0, &Enter, NULL);
```
krystian-hebel, line 1097: Does this include Enter on numpad? I always find it annoying when it doesn't work.

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1039,6 +1032,26 @@ PrintSolStrings (
...
+**/
+VOID
+EFIAPI
+RefreshLogo (
+  IN EFI_EVENT    Event,
+  IN VOID         *Context
+  )
+{
+  gBS->CloseEvent (Event);
+  gBS->Stall (100 * 1000);
```
krystian-hebel: Can this be moved after screen is redrawn?

Why is this delay even needed in the first place? Just to show 100% on progress bar for a while?

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1039,6 +1032,26 @@ PrintSolStrings (
   }
 }
 
+/**
+  Refresh the logo on ReadyToBoto event. It will clear the screen from strings
```
krystian-hebel:
```suggestion
  Refresh the logo on ReadyToBoot event. It will clear the screen from strings

```

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -709,6 +702,7 @@ PlatformBootManagerBeforeConsole (
   //
   FilterAndProcess (&gEfiPciRootBridgeIoProtocolGuid, NULL, Connect);
 
+  PlatformConsoleInit ();
```
krystian-hebel, line 705: Not really related to this change, but have you tested if output still works if console is initialized before `EfiBootManagerDispatchDeferredImages()`?

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
miczyg1: Well, haven't checked :) But probably yes.

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
miczyg1: PlatformConsoleInit simply adds a PCI UART, so it will work as soon as Root Bridge IO is connected

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
miczyg1: >  Just to show 100% on progress bar for a while?

Sort of... We may get rid of that I think

miczyg1:
```suggestion
```

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
krystian-hebel: Are you sure about that? I thought it also enabled stuff that depends on GOP, which in turn may depend on OptionROM.

miczyg1: Yes, I am sure: https://github.com/Dasharo/edk2/blob/dasharo/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformConsole.c#L538

It handles only PS2 keybaord and Intel PCI Serial ports

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
miczyg1: Yup, NUM ENTER also works (at least with USB KBD)

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1039,6 +1032,26 @@ PrintSolStrings (
...
+**/
+VOID
+EFIAPI
+RefreshLogo (
+  IN EFI_EVENT    Event,
+  IN VOID         *Context
+  )
+{
+  gBS->CloseEvent (Event);
+  gST->ConOut->ClearScreen (gST->ConOut);
```
mkopec, line 1051: This causes a short flcker before the logo is displayed again. Would be nice if we could do this without flickering

mkopec (approved):

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
miczyg1: You are asking for too much. Flickering was also present earlier if you did not press any button.

mkopec: Ok, it was more of a "Would Be Nice" thing than a request for change.

PR #64, opened 2023-07-18
SecurityPkg/Tcg/Opal/OpalPassword: check for both BLOCK_IO and BLOCK_IO2

miczyg1:


SecurityPkg/Tcg/Opal/OpalPassword/OpalDriver.c
```diff
@@ -2793,43 +2802,70 @@ OpalEfiDriverBindingStart(
...
+    BlkIo = NULL;
+
   if (EFI_ERROR(Status)) {
-    //
-    // Block_IO not supported on handle
-    //
-    if(Status == EFI_UNSUPPORTED) {
-      BlkIo = NULL;
-    } else {
+    Status2 = gBS->OpenProtocol(
```
mkopec, line 2810: Should EFI_BLOCK_IO2_PROTOCOL not take priority over EFI_BLOCK_IO_PROTOCOL?

mkopec (commented):
One question from me

SecurityPkg/Tcg/Opal/OpalPassword/OpalDriver.c
miczyg1: UEFI spec does not specify which protocol has priority. I have checked all drivers that produce BLOCK_IO or BLOCK_IO2 and it looks like all drivers produce both, so it shouldn't matter. All we need the BLOCK_IO for is the MediaId, which shouldn't be different depending on BLOCK_IO or BLOCK_IO2 @mkopec

mkopec: Ok, thanks.

mkopec (approved):

PR #72, opened 2023-09-02
Add runtime control over OptionROM loading

SergiiDmytruk:


UefiPayloadPkg/PciPlatformDxe/PciPlatformDxe.c
```diff
@@ -43,6 +45,88 @@ PciPlatformPrepController(
...
+  if (!EFI_ERROR (Status)) {
+      LoadOptionRom = IS_PCI_DISPLAY (&PciConfHeader);
+  }
+
+  gBS->CloseProtocol (
+      PciHandle,
+      &gEfiPciIoProtocolGuid,
+      PciIo,
+      PciHandle
+      );
```
miczyg1: We are not doing OpenProtocol, just HandleProtocol, so we shouldn't close the protocol AFAIK @SergiiDmytruk

miczyg1 (changes requested):

UefiPayloadPkg/PciPlatformDxe/PciPlatformDxe.c
SergiiDmytruk: Looks like you're right.  I wasn't sure about it, `PciGetPciRom()` does call `CloseProtocol()` after `HandleProtocol()` and I forgot to check with the spec.

miczyg1: I have double checked the spec and it lookes like OpenProtocol is an extended version of HandleProtocol. The difference is that HandleProtocol does not record anywhere if some agent is consuming a protocol. OpenProtocol does record which agent is cnsuming the protocol in a handle database and in such case it CloseProtocol should be required. So we can either:

1. Drop the CloseProtocol, OR
2. Replace HandleProtocol calls with OpenProtocol with attribute EFI_OPEN_PROTOCOL_BY_HANDLE_PROTOCOL

miczyg1: OK, I see you went with option 1. Closing discussion

miczyg1 (approved):

miczyg1:
Two more suggestions to address here: https://github.com/Dasharo/DasharoModulePkg/pull/26
Before this PR can be merged

PR #79, opened 2023-09-20
pre-commit: add config

macpijan:


macpijan:
We cannot add much more checks right now without fixing thousands of problems in edk2 codebase

miczyg1 (approved):

PR #82, opened 2023-09-25
UefiPayloadPkg/UefiPayloadPkg.dsc: Decrease s76 ec console verbosity

mkopec:
Make the EC console behave as a regular serial console.


mkopec:
@miczyg1 ping

macpijan (approved):
I have verified that with this patch (and edk2 serial enabled) I was able to use the device in automated tests.

miczyg1:
A good solution would be to AND the SYSTEM76_EC_LOGGING with `TARGET ==  DEBUG` so that when debug build is used, the EC console may act as debugging console as well.

miczyg1:
@mkopec added a PR for that: https://github.com/Dasharo/edk2/pull/84

PR #86, opened 2023-09-28
UefiPayloadPkg/Library/PlatformBootManagerLib: Save SMBIOS data in va…

miczyg1:
…riables


miczyg1:
Works as expected. In case the CBFS field are removed, the variables work as fallback and the DMI is populated in coreboot correctly with the previous values. Of course if one puts new CBFS files the current SN and UUID are overridden.

mkopec (approved):

PR #88, opened 2023-10-02
LaptopBatteryLib/LaptopBatteryLib.c: fix EC reg32 read

mkopec:
fixes the battery popup always showing 0% battery and blocking boot.

### DIFF
--- a/UefiPayloadPkg/Library/LaptopBatteryLib/LaptopBatteryLib.c
+++ b/UefiPayloadPkg/Library/LaptopBatteryLib/LaptopBatteryLib.c
@@ -180,6 +180,8 @@ EcReadReg32 (
   if (!Data32)
     return RETURN_INVALID_PARAMETER;
 
+  *Data32 = 0;
+
   for (Index = 0; Index < 4; Index++) {
     Status = EcReadReg(Reg + Index, &Data8[Index]);
 


miczyg1:
@mkopec https://github.com/Dasharo/coreboot/blob/clevo/develop/src/ec/system76/ec/acpi/ec_ram.asl#L21
...
All values are reported as 32bit, so what the heck? This implementation was based on clevo/develop branch back then.

If this is 16bit field, OS should not access it as 32bit right?

mkopec:
@miczyg1 here's the EC side implementation: https://github.com/Dasharo/ec/blob/d1dcb90227e4ee479f6d4657cb93f6ae014ccd05/src/board/system76/common/acpi.c#L121

the two highest bytes would always return 0 so it would pretty much always return 0%.

miczyg1:
> @miczyg1 here's the EC side implementation: https://github.com/Dasharo/ec/blob/d1dcb90227e4ee479f6d4657cb93f6ae014ccd05/src/board/system76/common/acpi.c#L121
> 
> the two highest bytes would always return 0 so it would pretty much always return 0%.

If the highest two bytes are zeros, then it should still work, because lower two bytes are meaningful.

I.e. what is the difference between:
1. 32bit 0x00001234
2. 16bit 0x1234

?? None. The only difference is the division precision: https://github.com/Dasharo/edk2/pull/88/files#diff-de4f0756b0ad7f1c8b395594265b6494afff3bd90ca79127cd13170e2dfe56aaR299

mkopec:
I don't know why ACPI says one thing and the EC something else, but before this change the battery level was always 0% and now it works reliably.

miczyg1:
> I don't know why ACPI says one thing and the EC something else, but before this change the battery level was always 0% and now it works reliably.

If you don't know, it means that the problem is somewhere else and you should investigate it a bit further. Your changes basically skip the two higher bytes and nothing else. I can't see how does it fix the problem. Maybe the higher two bytes are not zeros?

mkopec:
logs from EC with this change:

```
[13:52:04.841] acpi_read 03 = 01
[13:52:11.562] acpi_read 10 = 04
[13:52:11.566] acpi_read 10 = 04
[13:52:11.575] acpi_read 1A = B9
[13:52:11.583] acpi_read 1B = 26
[13:52:11.587] acpi_read 2E = 0D
[13:52:11.595] acpi_read 2F = 23
```

logs before:

```
[13:57:04.378] acpi_read 03 = 01
[13:57:13.827] acpi_read 10 = 04
[13:57:13.835] acpi_read 10 = 04
[13:57:13.840] acpi_read 1A = B9
[13:57:13.848] acpi_read 1B = 26
[13:57:13.856] acpi_read 1C = 00
[13:57:13.860] acpi_read 1D = 00
[13:57:13.868] acpi_read 2E = AB
[13:57:13.876] acpi_read 2F = 22
[13:57:13.880] acpi_read 30 = 00
[13:57:13.889] acpi_read 31 = 00
```

miczyg1:
> [13:57:13.840] acpi_read 1A = B9
> [13:57:13.848] acpi_read 1B = 26
> [13:57:13.856] acpi_read 1C = 00
> [13:57:13.860] acpi_read 1D = 00

As you can see, your changes should not matter, because high bytes are zeros. Maybe there is a problem in the math here: https://github.com/Dasharo/edk2/pull/88/files#diff-de4f0756b0ad7f1c8b395594265b6494afff3bd90ca79127cd13170e2dfe56aaR299

Check the dividend and divisor and the final result.

mkopec:
from edk2 before this change

```
Battery full capacity: 1985621945
Battery remaining capacity: 9084
```

mkopec:
@miczyg1 so it looks like the argument to `EcReadReg32` was not initialized anywhere, and we `or` it with the value from the EC...

funnily enough when trying to debug, on subsequent calls to `LaptopGetBatteryCapacity` I would get a valid value back. Oh well.

i guess reading the entire 32bits can stay in this case.

miczyg1:
> from edk2 before this change
> 
> ```
> Battery full capacity: 1985621945
> Battery remaining capacity: 9084
> ```

What these values refer to?
I can assume `Battery remaining capacity: 9084` is RemainingCap. Is `Battery full capacity: 1985621945` the `LastFullChargeCap`?

mkopec:
@miczyg1 see comment above and updated patch.

miczyg1:
> @miczyg1 so it looks like the argument to `EcReadReg32` was not initialized anywhere, and we `or` it with the value from the EC...
> 
> funnily enough when trying to debug, on subsequent calls to `LaptopGetBatteryCapacity` I would get a valid value back. Oh well.
> 
> i guess reading the entire 32bits can stay in this case.

But why is the initialization even needed?  We always read 4 bytes, so the whole size of variable is rewritten. If any error occurs during the read, the variable is not used...

miczyg1:
> @miczyg1 so it looks like the argument to `EcReadReg32` was not initialized anywhere, and we `or` it with the value from the EC...
> 
> funnily enough when trying to debug, on subsequent calls to `LaptopGetBatteryCapacity` I would get a valid value back. Oh well.
> 
> i guess reading the entire 32bits can stay in this case.

Ahh right, it is because of the OR operation....

miczyg1 (approved):

PR #89, opened 2023-10-02
MdeModulePkg/Universal/Acpi/AcpiTableDxe: copy XFirmwareCtl unconditionally

mkopec:
Since coreboot commit https://github.com/Dasharo/coreboot/commit/a07b09ab71d04b870b9413b7e075a2369f59fd0a , coreboot looks for FACS at x_firmware_ctl when resuming from s3. EDK2 only allowed the 64 bit or 32 bit FACS address to be set at once and would zero out the other. This means that when FACS resides below 4G, x_firmware_ctl would be set zero and coreboot would fail to find the resume vector on wakeup from S3.

So just copy XFirmwareCtl from coreboot always.


mkopec:
Rebased and signed.

mkopec:
@miczyg1 Turns out this is probably needed on laptops too, at the moment S3 resume does not work.

mkopec:
Confirmed to fix S3 on NV4x ADL.

miczyg1 (approved):

PR #90, opened 2023-10-08
.github/workflows/build.yml: add

macpijan:


macpijan:
The problem is coreboot container uses limited DASH shell by default, not BASH. It has no source command and cannot interpret correctly edksetup.sh script. In github actions there is no way to override entrypoint of the container it seems, so I needed to workaround this via docker run

.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
...
+        fi
+
+    - name: Check DEBUG build Artifacts
+      run: |
+        if [ -f "Build/OvmfX64/DEBUG_GCC5/FV/OVMF_CODE.fd" ] && [ -f "Build/OvmfX64/DEBUG_GCC5/FV/OVMF_VARS.fd" ]; then
+          echo "DEBUG build successful. OVMF firmware image files found."
+        else
+          echo "DEBUG build failed. OVMF firmware image files not found."
+          exit 1
+        fi
```
miczyg1, line 51: No artifacts uploading? I would at least expect the RELEASE artifacts to be uploaded in case anyone needs to test something on QEMU and not ask others to compile it...

miczyg1 (changes requested):

.github/workflows/build.yml
macpijan: @miczyg1 Added, see this run: https://github.com/Dasharo/edk2/actions/runs/6469542116

macpijan: @miczyg1 Please resolve if this is enough

miczyg1 (approved):

PR #94, opened 2023-10-17
Autoboot timeout fix

filipleple:


MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
```diff
@@ -1093,6 +1094,15 @@ BootMaintCallback (
 
   Private        = BMM_CALLBACK_DATA_FROM_THIS (This);
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_FIRMWARE) {
```
miczyg1:
```suggestion
  if (Action == EFI_BROWSER_ACTION_RESET_TO_DEFAULT) {
```

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
```diff
@@ -1093,6 +1094,15 @@ BootMaintCallback (
 
   Private        = BMM_CALLBACK_DATA_FROM_THIS (This);
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_FIRMWARE) {
+    //gotta find something like Private.BmmDefaultValues.BootTimeOut;
+    CurrentFakeNVMap->BootTimeOut = defaultValue
```
miczyg1: You have to filter for Auto Boot TImeout QuestionID. For other it has to return EFI_UNSUPPORTED.

The value is being set using the `IN        EFI_IFR_TYPE_VALUE                     *Value,` parameter. So you should probably do `Value->u16 = PcdGet16 (PcdPlatformBootTimeOut);` If this doesn't work, then `CurrentFakeNVMap->BootTimeOut =  PcdGet16 (PcdPlatformBootTimeOut);` will probably be the way

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
```diff
@@ -1093,6 +1094,15 @@ BootMaintCallback (
 
   Private        = BMM_CALLBACK_DATA_FROM_THIS (This);
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_FIRMWARE) {
+    //gotta find something like Private.BmmDefaultValues.BootTimeOut;
+    CurrentFakeNVMap->BootTimeOut = defaultValue
+
+    // Update the value in the HII database
+    HiiSetBrowserData (&mBootMaintGuid, mBootMaintStorageName, sizeof (BMM_FAKE_NV_DATA), (UINT8 *) CurrentFakeNVMap, NULL);
```
miczyg1: This function is called at the end of BootMaintCallback . The value will be saved when the save hotkey is pressed (also handled in separate action most likely).

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
```diff
@@ -1093,6 +1094,15 @@ BootMaintCallback (
 
   Private        = BMM_CALLBACK_DATA_FROM_THIS (This);
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_FIRMWARE) {
+    //gotta find something like Private.BmmDefaultValues.BootTimeOut;
+    CurrentFakeNVMap->BootTimeOut = defaultValue
+
+    // Update the value in the HII database
+    HiiSetBrowserData (&mBootMaintGuid, mBootMaintStorageName, sizeof (BMM_FAKE_NV_DATA), (UINT8 *) CurrentFakeNVMap, NULL);
+    return EFI_SUCCESS;
```
miczyg1: No need to return here as the other IF branches will protect fro executing any other code, i.e. if `Action` is `EFI_BROWSER_ACTION_RESET_TO_DEFAULT`, it will not have any other value in given call of BootMaintCallback unless you explicitly change it.

miczyg1 (changes requested):

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
filipleple: @miczyg1 It won't compile with EFI_BROWSER_ACTION_RESET_TO_DEFAULT. Does that mean I'm missing something or is EFI_BROWSER_ACTION_REQUEST_RESET, as suggested by the compiler, valid?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
filipleple: @miczyg1 both `Value->u16 = PcdGet16 (PcdPlatformBootTimeOut);` and `CurrentFakeNVMap->BootTimeOut =  PcdGet16 (PcdPlatformBootTimeOut);` still cause the timeout value to be restored to `0`. Could the default value returned by `PcdGet16` really be `0`?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
miczyg1: Actually the proper define for that is `EFI_BROWSER_ACTION_DEFAULT_STANDARD` and `EFI_BROWSER_ACTION_DEFAULT_MANUFACTURING`

miczyg1:
I have prepared an alternative approach how to add default value support for dynamically generated IFR opcodes:
https://github.com/Dasharo/edk2/pull/96

It looks like the last parameter of HiiCreateNumericOpCode can take any Opcode which is appended to the primary opcode generated by HiiCreateNumericOpCode. In this case I have created a Default value opcode with HiiCreateDefaultOpCode

miczyg1:
Merged the https://github.com/Dasharo/edk2/pull/96 so this PR is obsolete.

PR #95, opened 2023-10-17
Reset to default opt

miczyg1:


mkopec:
This can probably be replaced with https://github.com/Dasharo/edk2/pull/109 which fixes the F9 hotkey behavior.

PR #98, opened 2023-10-20
Reject negative autoboot timeout fix

### LINKED ISSUE #505: Auto Boot Time-out do not show error for `-` value
**Dasharo version**
v1.1.0

**Dasharo variant**
protectli_vp4670

**Affected component(s) or functionality**

`Auto Boot Time-out` in Dasharo Menu

**Brief summary**

If we press a key with a value that is not a digit, two red exclamation marks `!!` appear at the bottom of the window. It is still waiting for a numerical value.

If we press `-`, it exits the value entry, does not wait for the number to be entered.
In addition, I do not ask whether to save the value, even if we have previously changed it to another numerical value.

**How reproducible**

Always

**How to reproduce**

Press `-` where you enter the `Auto Boot Time-out` value. 

**Expected behavior**

If we press `-`, two red exclamation marks `!!` appear at the bottom of the window. It is still waiting for a numerical value. 

**Actual behavior**

If we press `-`, it exits the value entry, does not wait for the number to be entered.
In addition, I do not ask whether to save the value, even if we have previously changed it to another numerical value.

**Screenshots**

image
If we press a key with a value that is not a digit (and not `-` because of a bug), two red exclamation marks `!!` appear at the bottom of the window. It is still waiting for a numerical value.

filipleple:


MdeModulePkg/Universal/DisplayEngineDxe/InputHandler.c
```diff
@@ -619,7 +619,7 @@ GetNumericInput (
   if ((Question->OpCode->OpCode == EFI_IFR_NUMERIC_OP) && (NumericOp != NULL)) {
     if ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_HEX){
       HexInput = TRUE;
-    } else if ((NumericOp->Flags & EFI_IFR_DISPLAY) == 0){
+    } else if ( ((NumericOp->Flags & EFI_IFR_DISPLAY) == 0) || ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_DEC) ){
```
miczyg1: This shouldn't be needed. See next comment.

MdeModulePkg/Universal/DisplayEngineDxe/InputHandler.c
```diff
@@ -752,6 +752,11 @@ GetNumericInput (
         // In Manual input mode, check whether input the negative flag.
         //
         if (Key.UnicodeChar == '-') {
+          if ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_DEC){
```
miczyg1: What I would do here is simply check for UINT and ManualInput, because this fragment handles two cases:

1. ManualInput and type is INT. `+` and `-` need special treat here for signed integers, and this is what is currently made (before modifications).
2. If not ManualInput, then `+` and `-` is used to increase numeric by the value of `step` attribute. Not all numerics have step attribute. Some have step attribute set to 0, which means only ManualInput.

What we need to do here is most likely something like this:

```C
TheKey2:
    switch (Key.UnicodeChar) {

    case '+':
    case '-':
+     if (ManualInput && ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_DEC)) {
+       UpdateStatusBar(INPUT_ERROR, TRUE);
+       break;
+     }
      if (ManualInput && IntInput) {
        //
        // In Manual input mode, check whether input the negative flag.
        //
        if (Key.UnicodeChar == '-') {
          if (Negative) {
            break;
          }
          Negative = TRUE;
          PrintCharAt (Column++, Row, Key.UnicodeChar);
        }
      } else {
        if (Key.UnicodeChar == '+') {
          Key.ScanCode = SCAN_RIGHT;
        } else {
          Key.ScanCode = SCAN_LEFT;
        }
        Key.UnicodeChar = CHAR_NULL;
        goto TheKey2;
      }
      break;
```

This should solve it right?

miczyg1 (changes requested):

MdeModulePkg/Library/CustomizedDisplayLib/CustomizedDisplayLib.uni
```diff
@@ -43,8 +43,8 @@
                                        #language fr-FR  "<Spacebar>Bascule la Case de pointage"
 #string NV_UPDATE_MESSAGE              #language en-US  "Configuration changed"
                                        #language fr-FR  "Configuration changed"
-#string INPUT_ERROR_MESSAGE            #language en-US  "!!"
-                                       #language fr-FR  "!!"
+#string INPUT_ERROR_MESSAGE            #language en-US  "Invalid input value"
+                                       #language fr-FR  "Invalid input value"
```
miczyg1:
```suggestion
#string INPUT_ERROR_MESSAGE            #language en-US  "Invalid input value!!"
                                       #language fr-FR  "Invalid input value!!"
```

MdeModulePkg/Universal/DisplayEngineDxe/InputHandler.c
miczyg1: Above code snippet tested and QEMU and behaves as intended

MdeModulePkg/Universal/DisplayEngineDxe/InputHandler.c
```diff
@@ -619,7 +619,7 @@ GetNumericInput (
   if ((Question->OpCode->OpCode == EFI_IFR_NUMERIC_OP) && (NumericOp != NULL)) {
     if ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_HEX){
       HexInput = TRUE;
-    } else if ((NumericOp->Flags & EFI_IFR_DISPLAY) == 0){
+    } else if ( ((NumericOp->Flags & EFI_IFR_DISPLAY) == 0)  ){
```
miczyg1:
```suggestion
    } else if ((NumericOp->Flags & EFI_IFR_DISPLAY) == 0){
```

MdeModulePkg/Universal/DisplayEngineDxe/InputHandler.c
```diff
@@ -747,6 +747,10 @@ GetNumericInput (
 
     case '+':
     case '-':
+    if (ManualInput && ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_DEC)) {
+       UpdateStatusBar(INPUT_ERROR, TRUE);
+       break;
+     }
```
miczyg1:
```suggestion
      if (ManualInput && ((NumericOp->Flags & EFI_IFR_DISPLAY) == EFI_IFR_DISPLAY_UINT_DEC)) {
        UpdateStatusBar(INPUT_ERROR, TRUE);
        break;
      }
```

miczyg1 (dismissed):
@filipleple something is off with your editor or the code autoformatting does not work too well.

MdeModulePkg/Universal/DisplayEngineDxe/InputHandler.c
filipleple: right, it always gives me trouble if the indents are not 4*n spaces, will fix

filipleple:
@miczyg1 all fixes applied, your suggested solution is implemented

mkopec:
Looks good, looks like all comments in review were addressed, marking them as resolved.

@filipleple In general, please remember to add link to the fixing commit in response to review comments, according to our guidelines

mkopec (approved):

PR #99, opened 2023-10-20
UefiPayloadPkg: Use DxeRuntimeDebugLibSerialPort in DXE_RUNTIME_DRIVER

miczyg1:
Prevent debugging on serial port (whether physical or cbmem console) at runtime by using the DxeRuntimeDebugLibSerialPort library as DebugLib. It will stop calling SerialPortWrite if EFI switches to runtime and avoid access to cbmem CONSOLE buffer which is neither marked as runtime code nor data. Solves the issue with Xen backtrace on EFI reset system runtime service: https://github.com/Dasharo/dasharo-issues/issues/488#issuecomment-1772758851


miczyg1:
UNTESTED

miczyg1:
Tried to reproduce the problem on QubesOS, but shutdown didn't trigger the crash (tried few times). The cbmem log information is not lost after this change and its content is the same as before the change.

PR #100, opened 2023-10-20
UefiPayloadPkg: Enable TCG MOR

miczyg1:


miczyg1:
Currently the MOR lock variable is created with size 0 and attributes 0, because we have no SMM protection over it. So basically OS deems MOR to be unsupported.

We would still need to find a way how to handle MOR lock via SMM in coreboot.

macpijan:
@miczyg1 Is it still draft, or we want it merged? What benefit it provides?

miczyg1:
> @miczyg1 Is it still draft, or we want it merged? What benefit it provides?

It is draft on purpose. Currently it has no way of working properly (need to design a way to protect MOR variables somehow, can't use EDK2 code, because it relies on EDK2 SMM, while SMM is initialized in coreboot only). The benefit was working Windows 11 VBS, hopefully.

PR #101, opened 2023-10-23
Add support for iPXE and make Dasharo netwok boot menu work

### LINKED ISSUE #543: iPXE with Dasharo network boot menu shall be selectable from the boot menu, only if the Network Boot is enabled

pietrushnic:
Fixes: https://github.com/Dasharo/dasharo-issues/issues/543

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - dasharo
     tags:
-      - 'dasharo_qemu_v*'
+      - 'qemu_q35_v*'
   pull_request:
     branches:
       - dasharo
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,24 @@ jobs:
         git checkout 3323ed481d35096fb6a7eae7b49f35eff00f86cf && \
         cd -
 
+    - name: Clone iPXE Repository
+      run: |
+        git clone https://git.ipxe.org/ipxe.git && \
+        cd ipxe && \
+        git checkout 4bffe0f0d9d0e1496ae5cfb7579e813277c29b0f && \
+        sed -i 's|//#define\s*IMAGE_SCRIPT.*|#define IMAGE_SCRIPT|' "src/config/general.h" && \
+        sed -i 's|.*DOWNLOAD_PROTO_HTTPS|#define DOWNLOAD_PROTO_HTTPS|g'  "src/config/general.h" && \
+        wget https://raw.githubusercontent.com/Dasharo/dasharo-blobs/main/dasharo/dasharo.ipxe && \
+        cd -
+
+    - name: Build iPXE
+      run: |
+        docker run --rm -i -v $PWD/ipxe:/home/coreboot/ipxe:rw \
+          -v $PWD/.github:/home/coreboot/ipxe/.github \
+          -u $(id -u):$(id -g) -w /home/coreboot/ipxe \
+          coreboot/coreboot-sdk:2021-09-23_b0d87f753c \
+         ./.github/scripts/build-ipxe.sh
+
     - name: Build OVMF Firmware Image
       run: |
         docker run --rm -i -v $PWD:/home/coreboot/coreboot:rw \
--- a/OvmfPkg/OvmfPkg.dec
+++ b/OvmfPkg/OvmfPkg.dec
@@ -256,10 +256,6 @@
   ## Number of page frames to use for storing grant table entries.
   gUefiOvmfPkgTokenSpaceGuid.PcdXenGrantFrames|4|UINT32|0x33
 
-  ## FFS filename to find the ipxe application.
-  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile|{ 0xC7, 0x53, 0x86, 0xb6, 0xA1, 0xEE, 0x35, 0x44, 0xA1, 0x99, 0xA4, 0x4F, 0x59, 0xE4, 0x47, 0x6C }|VOID*|0x10000006
-  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"|VOID*|0x10000007
-
 [PcdsDynamic, PcdsDynamicEx]
   gUefiOvmfPkgTokenSpaceGuid.PcdEmuVariableEvent|0|UINT64|2
   gUefiOvmfPkgTokenSpaceGuid.PcdOvmfFlashVariablesEnable|FALSE|BOOLEAN|0x10
--- a/OvmfPkg/OvmfPkg.dec
+++ b/OvmfPkg/OvmfPkg.dec
@@ -256,6 +256,10 @@
   ## Number of page frames to use for storing grant table entries.
   gUefiOvmfPkgTokenSpaceGuid.PcdXenGrantFrames|4|UINT32|0x33
 
+  ## FFS filename to find the ipxe application.
+  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile|{ 0xC7, 0x53, 0x86, 0xb6, 0xA1, 0xEE, 0x35, 0x44, 0xA1, 0x99, 0xA4, 0x4F, 0x59, 0xE4, 0x47, 0x6C }|VOID*|0x10000006
+  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE"|VOID*|0x10000007
+
 [PcdsDynamic, PcdsDynamicEx]
   gUefiOvmfPkgTokenSpaceGuid.PcdEmuVariableEvent|0|UINT64|2
   gUefiOvmfPkgTokenSpaceGuid.PcdOvmfFlashVariablesEnable|FALSE|BOOLEAN|0x10
--- a/OvmfPkg/OvmfPkg.dec
+++ b/OvmfPkg/OvmfPkg.dec
@@ -258,7 +258,7 @@
 
   ## FFS filename to find the ipxe application.
   gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile|{ 0xC7, 0x53, 0x86, 0xb6, 0xA1, 0xEE, 0x35, 0x44, 0xA1, 0x99, 0xA4, 0x4F, 0x59, 0xE4, 0x47, 0x6C }|VOID*|0x10000006
-  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE"|VOID*|0x10000007
+  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"|VOID*|0x10000007
 
 [PcdsDynamic, PcdsDynamicEx]
   gUefiOvmfPkgTokenSpaceGuid.PcdEmuVariableEvent|0|UINT64|2
--- a/.github/scripts/build-ipxe.sh
+++ b/.github/scripts/build-ipxe.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+export CROSS_COMPILE="x86_64-elf-"
+make -C src bin-x86_64-efi-sb/ipxe.efi EMBED=$PWD/dasharo.ipxe BUILD_ID_CMD="echo 0x1234567890" \
+  EXTRA_CFLAGS="-Wno-address-of-packed-member  -m64  -fuse-ld=bfd \
+  -Wl,--build-id=none -fno-delete-null-pointer-checks -Wlogical-op -march=nocona \
+  -malign-data=abi -mcmodel=large -mno-red-zone -fno-pic"
+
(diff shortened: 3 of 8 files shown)


OvmfPkg/OvmfPkg.dec
```diff
@@ -256,6 +256,10 @@
   ## Number of page frames to use for storing grant table entries.
   gUefiOvmfPkgTokenSpaceGuid.PcdXenGrantFrames|4|UINT32|0x33
 
+  ## FFS filename to find the ipxe application.
+  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile|{ 0xC7, 0x53, 0x86, 0xb6, 0xA1, 0xEE, 0x35, 0x44, 0xA1, 0x99, 0xA4, 0x4F, 0x59, 0xE4, 0x47, 0x6C }|VOID*|0x10000006
+  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"|VOID*|0x10000007
```
miczyg1: You already import UefiPayloadPkg/UefiPayloadPkg.dec in OvmfPkg/Library/PlatformBootManagerLib/PlatformBootManagerLib.inf, so these definitions are redundant

miczyg1 (changes requested):

OvmfPkg/OvmfPkg.dec
pietrushnic: @miczyg1 yes, but I would like to use a different value. Shall I fall back to the built time parameter to avoid this redundancy?

pietrushnic: Fixed in aa7fdee951de

miczyg1: Then, simply put `gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"` to the DSC file

macpijan (commented):
No actual code review on my side. I can just confirm that with the build I took from CI I could run these two tests using iPXE: https://github.com/Dasharo/open-source-firmware-validation/pull/63/commits/2b55fd5f45aec09b4d2c764da7507da0d8a4f45d

OvmfPkg/OvmfPkg.dec
pietrushnic: @miczyg1 I already made it runtime, as shown in the commit. The runtime method is good for all other builds and not for QEMU? If that is the case, then why?

pietrushnic:
@macpijan do you think we need a detailed review? Who can do that?

OvmfPkg/OvmfPkg.dec
miczyg1: And why would you need to change it for QEMU? I don't see such a need. In case of UEFIPaylaod we need to override the iPXE boot option name for Protecli builds and leave `iPXE Network Boot` for non-Protectli builds.

You can always pass the `--pcd gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName=value` to the build command to override it. Just put `gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"` to the DSC file and that's all (runtime overriding will still work...) @pietrushnic

macpijan:
@pietrushnic It looks to me that @miczyg1 already looked into that and when we resolve open thread I think we are good to move forward.

OvmfPkg/OvmfPkg.dec
pietrushnic: > And why would you need to change it for QEMU? I don't see such a need. In case of UEFIPaylaod we need to override the iPXE boot option name for Protecli builds and leave iPXE Network Boot for non-Protectli builds.

The only reason I see now is consistency across supported platforms, but that can be changed in platform configuration in OSFV. What I'm worried about most in this review is that we are getting to that realization after so many hours spent beating a dead horse. This is not the first time. We have to think about how to change our feedback culture.

> You can always pass the --pcd gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName=value to the build command to override it. Just put gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot" to the DSC file and that's all (runtime overriding will still work...)

@miczyg1 I don't get that point. This is precisely what I did. First, you complained about a duplicated line, so I removed it and added as a build-time parameter https://github.com/Dasharo/edk2/commit/aa7fdee951deb45f03162d143a30eae784b7fd1f
Also changed documentation to accommodate that: https://github.com/Dasharo/docs/pull/684/commits/a03b26d9cf79db5ea4c09598c85e1eee0b0074d2

macpijan: I'm not sure what the conclusion is and will probably not help on how to implement it in the edk2 build system.

I can add just one thing, that in general it sounds like it would make sense that the Ovmf target have similar level of configurability as the Uefipayload.
For instance, I might want to have another config for QEMU, that would build me the "Protectli" (or any other) variant of the build, to test some features in QEMU - such as the customized network menu in this particular case.

pietrushnic: @macpijan OK, I will change the documentation tests and CI scripts to comply with @miczyg1 request. I just wanted to highlight I'm doing this a second or third time, changing essentially the same code because, from the beginning, we could not express a good enough vision of how it should be done.

pietrushnic: @miczyg1 fixes:
* OSFV: https://github.com/Dasharo/open-source-firmware-validation/pull/71/commits/03f0b460c7bef7f8322b4259ce726c53b951f095
* edk2 CI changes: 3c61fe914bc2
* documentation correction: https://github.com/Dasharo/docs/pull/684/commits/7a66277e2de7eda8b4844bed26891507e4f195e5

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - dasharo
     tags:
-      - 'dasharo_qemu_v*'
+      - 'qemu_q35_v*'
```
pietrushnic, line 8: @macpijan since you created edk2 CI, it would be great if you could review the changes in this file.

macpijan: @pietrushnic I'm fine with these

OvmfPkg/OvmfPkg.dec
pietrushnic: @miczyg1 ?

pietrushnic: @miczyg1 ?

pietrushnic:
@miczyg1 I opened #104 to continue work related with UEFI Secure Boot.

OvmfPkg/OvmfPkg.dec
miczyg1: > @miczyg1 I don't get that point. This is precisely what I did. First, you complained about a duplicated line, so I removed it and added as a build-time parameter https://github.com/Dasharo/edk2/commit/aa7fdee951deb45f03162d143a30eae784b7fd1f

Oh no no no. At first you added that to DEC file which I complained on. Then I asked to add that to DSC if you want overrides. DEC != DSC with all due respect.

miczyg1: I don't see the `gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"` defined in OvmfPkgX64.dsc. Still up for that one or not?

pietrushnic: @miczyg1 this is my understanding of our conversation:

> At first you added that to DEC file which I complained on. 

Correct. This is the first comment in this thread. You did not suggest anything related to DSC in your first comment. I explained why I did things in such a way and asked for possible solutions using runtime parameters. 

I didn't get any reply to my question. That's why I implemented aa7fdee951deb45f03162d143a30eae784b7fd1f, which addresses your first comment because changes from DEC are removed. My commit uses built-time parameters as coreboot does.

Then you answered my question with:

> Then, simply put gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot" to the DSC file

The goal of this PR is as the title said add iPXE and make the Dasharo network boot work with the goal of passing network boot tests. I was wrong in thinking I needed modification of this PCD from the beginning. What you seem to confirm with the next comment.

> And why would you need to change it for QEMU? I don't see such a need. In case of UEFIPaylaod we need to override the iPXE boot option name for Protecli builds and leave iPXE Network Boot for non-Protectli builds.

What, IMHO, contradicts the DSC modifications suggested before, but is a valid concern from your side.

> You can always pass the --pcd gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName=value to the build command to override it. Just put gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot" to the DSC file and that's all (runtime overriding will still work...) @pietrushnic

I can do that without DSC modifications. I have `UefiPayloadPkg/UefiPayloadPkg.dec` and `gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName` in `./OvmfPkg/Library/PlatformBootManagerLib/PlatformBootManagerLib.inf`, what you can see in changes in this PR. This is enough to set up that PCD built-time.

Finally, after all that discussion, I realized I don't need those changes at all because I can change the OSFV QEMU platform config to cover the default situation, so I removed all the stuff I introduced. I did it here: https://github.com/Dasharo/edk2/pull/101#discussion_r1372386459

> Oh no no no. 
> DEC != DSC with all due respect.

Agree. Let's treat each with respect.

> I don't see the gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot" defined in OvmfPkgX64.dsc. Still up for that one or not?

Not.

Please let me know if you have any actionable tasks required to resolve this conversation.

pietrushnic: @miczyg1, we can reopen the discussion if needed in issues. For now, we have to push things forward. This PR does not affect anything other than OvmfPkg, so we should be safe merging that. Otherwise, we will never reach v0.1.0. Feel free to create issue if you think something should be improved I would be happy to implement fixes.

miczyg1: All good now. This is the correct state right now. FYI:

```
  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile|{ 0xC7, 0x53, 0x86, 0xb6, 0xA1, 0xEE, 0x35, 0x44, 0xA1, 0x99, 0xA4, 0x4F, 0x59, 0xE4, 0x47, 0x6C }|VOID*|0x10000006
  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"|VOID*|0x10000007
```

Was initially declared in OvmfPkg.dec, which was not necessary, because these definitions are already in UefiPayloadkPkg.dec, so UefiPayloadkPkg.dec can be included in the INF files that use the gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile and gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName (suggested in https://github.com/Dasharo/edk2/pull/101#discussion_r1370150335)

Then you asked to use different value: https://github.com/Dasharo/edk2/pull/101#discussion_r1370748755
So I suggested: https://github.com/Dasharo/edk2/pull/101#discussion_r1371310334 because `gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName` is the only PCD that make sense for value change. Putting `gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName` to DSC will allow to override it for example from the `build` command in CLI. However what you did instead [here](https://github.com/Dasharo/edk2/commit/aa7fdee951deb45f03162d143a30eae784b7fd1f#diff-d22f398fb5a433302a208a66be648fd91d04bcf941c80e748acd33e022dc51cbR19) is to override the value of the PCD with the very same default value set in UefiPaylodPkg.dec, but didn't put it into DSC, which I couldn't understand and wrote [this comment](https://github.com/Dasharo/edk2/pull/101#discussion_r1372017167).
Then you wrote [here](https://github.com/Dasharo/edk2/pull/101#discussion_r1372349983) that you `This is precisely what I did.`, but what I asked for was:

1. Remove from DEC (https://github.com/Dasharo/edk2/pull/101#discussion_r1370150335)

```
  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEFile|{ 0xC7, 0x53, 0x86, 0xb6, 0xA1, 0xEE, 0x35, 0x44, 0xA1, 0x99, 0xA4, 0x4F, 0x59, 0xE4, 0x47, 0x6C }|VOID*|0x10000006
  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"|VOID*|0x10000007
```

2. Add  `  gUefiPayloadPkgTokenSpaceGuid.PcdiPXEOptionName|L"iPXE Network Boot"` to DSC (https://github.com/Dasharo/edk2/pull/101#discussion_r1371310334)

And that's all.  DSC != DEC, so not "precisely what you did". I recall problems with runtime override in the `build` command line if some PCD (from DasharoModulePkg) was not defined in DSC, thus I suggested 2. to be sure.

tl;dr; all good

PR #102, opened 2023-10-24
New secure boot

miczyg1:
Developed during Qubes OS Summit 2023, and approved by Qubes OS UX expert :)


miczyg1:
Tested on MSI boards. Not sure yet how I can implement setting correct boot mode for QEMU to make OVMF testing possible.

miczyg1:
Accompanying documentation: https://github.com/Dasharo/docs/pull/685

macpijan:
@miczyg1 

> Not sure yet how I can implement setting correct boot mode for QEMU to make OVMF testing possible.

How to make sure of that?

We need this to work on the QEMU the same way it does on real hardware.
It will be a blocker for @pietrushnic  here: https://github.com/Dasharo/dasharo-issues/issues/538

pietrushnic:
Let's move this discussion to https://github.com/Dasharo/dasharo-issues/issues/538  - I already started to look at the problem yesterday from an edk2 perspective.

macpijan:
@miczyg1 Problem you have with the existing implementation is one problem to solve. Problem with newly introduced changes is another one. Let's keep track of both of them in the issue then. Because even if you solve the first one, we will still not be able to use QEMU SB with the most recent dasharo edk2 code.

PR #104, opened 2023-10-26
Dasharo UEFI Secure Boot support for OVMF

pietrushnic:

PR #106, opened 2023-10-26
WIP: Dasharo UEFI Secure Boot support in QEMU Q35

### LINKED ISSUE #550: Secure Boot Suite shall pass under QEMU

pietrushnic:
Fixes https://github.com/Dasharo/dasharo-issues/issues/550

Closes #104

### DIFF
--- a/.github/scripts/build-ipxe.sh
+++ b/.github/scripts/build-ipxe.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+export CROSS_COMPILE="x86_64-elf-"
+make -C src bin-x86_64-efi-sb/ipxe.efi EMBED=$PWD/dasharo.ipxe BUILD_ID_CMD="echo 0x1234567890" \
+  EXTRA_CFLAGS="-Wno-address-of-packed-member  -m64  -fuse-ld=bfd \
+  -Wl,--build-id=none -fno-delete-null-pointer-checks -Wlogical-op -march=nocona \
+  -malign-data=abi -mcmodel=large -mno-red-zone -fno-pic"
+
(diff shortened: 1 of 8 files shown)


macpijan:
@pietrushnic Was anything missing in this MR?

pietrushnic:
@macpijan this is not validated. Was stuck on OS installation from netboot.xyz, before I can pass the next test. I would focus on merging things related to iPXE. Here there is nothing more yet and I don't know if there will be if UEFi Secure Boot will work on #101 + miczyg's new UEFI Secure Boot.

macpijan:
@pietrushnic So I guess we do not need this anymore? SB works for me from the main branch (except the keys are reset by default, which is different than on real HW).

pietrushnic:
> @pietrushnic So I guess we do not need this anymore?

@macpijan most likely yes. So far I have not introduced any EDK II modifications related to secure boot, but we didn't test it comprehensively.

> except the keys are reset by default, which is different than on real HW

I don't understand that part. From what I see I can't enable secure boot if I do not reset keys which equals to provisioning default PK, db, dbx and KEK: https://docs.dasharo.com/dasharo-menu-docs/device-manager/#custom-mode-and-key-management)

macpijan:
@pietrushnic 

> From what I see I can't enable secure boot if I do not reset keys

On the QEMU you cannot, on the hardware you can. The initial behavior is different, that the default keys on the HW are already provisioned after flashing.

macpijan:
I could pass SB tests without these changes: https://github.com/Dasharo/open-source-firmware-validation/pull/118#issuecomment-1841613276

PR #107, opened 2023-11-02
OvmfPkg/OvmfPkg.dec: enable ESP scanning 

filipleple:


MdeModulePkg/MdeModulePkg.dec
```diff
@@ -2049,7 +2049,7 @@
   #   TRUE  - Create boot options for pre-installed OSes.<BR>
   #   FALSE - Do not create boot options for pre-installed OSes.
   # @Prompt Create boot options for pre-installed OSes.
-  gEfiMdeModulePkgTokenSpaceGuid.PcdCreatePreInstalledBootOptions|FALSE|BOOLEAN|0x0001007a
+  gEfiMdeModulePkgTokenSpaceGuid.PcdCreatePreInstalledBootOptions|TRUE|BOOLEAN|0x0001007a
```
macpijan: @filipleple We do not want to change the default value in this module.
We need to override this value just for OvmfPkg.

OvmfPkg/OvmfPkg.dec
```diff
@@ -309,3 +309,4 @@
   #  firmware contains a CSM (Compatibility Support Module).
   #
   gUefiOvmfPkgTokenSpaceGuid.PcdCsmEnable|FALSE|BOOLEAN|0x35
+  gEfiMdeModulePkgTokenSpaceGuid.PcdCreatePreInstalledBootOptions|TRUE|BOOLEAN|0x0001007a
```
macpijan: @filipleple It is already defined in `./MdeModulePkg.dec` - I'm not sure if we should define it also here. What is the problem if you do not define it here?

filipleple: @macpijan  Turned out to be overkill, removed in commit [bf80838](https://github.com/Dasharo/edk2/commit/bf80838ddff98bad5596cd1a3d43e254bb21313e)

macpijan: Thanks

macpijan (approved):

macpijan:
Applied on dasharo branch here: https://github.com/Dasharo/edk2/commit/1a54d5194452928bc97de197a658c9b46357d658

PR #110, opened 2023-11-09
SecureBootConfigDxe/SecureBootConfigImpl.c: add default for SB state

mkopec:
Fixes an issue where the setting would not be restored to the default value on pressing F9.


macpijan:
It works, but only when F9 is pressed in Secure Boot menu - still some improvement is likely needed here

mkopec:
@macpijan Fixed the Secure Boot setting not being restored and settings not being saved on reset - added a reminder to save (or discard) when selecting reset.

macpijan:
Thanks. Tested in QEMU as well.

macpijan (approved):

miczyg1:
Okay that seems to restore SB state. But what about keys? Are they restored to defaults? Also, if someone modified the keys and didn't enroll PK, setting SB state to enabled (if PcdSecureBootDefaultEnable is true/1) will fail.

PR #113, opened 2023-11-21
Add CI to check if UEFI DBX is up to date

mkopec:
Ideally this would be updatable via fwupd / ESRT, but until then, just check if the file is up to date.


miczyg1 (approved):

mkopec:
Huh, dbx is already updatable via fwupd and presumably Windows, it didn't appear for me a while ago but it seems they've fixed it

PR #115, opened 2024-01-03
MdeModulePkg/Application/UiApp: Customize front page

miczyg1:


miczyg1:
Wanted to give the front page a little better look and feel + RTC settings. Sample view from QEMU:

image

miczyg1:
The only problem I see with the current code is that EDK2 refreshes the banners on top and the hotkey help section on the bottom every second now. The refreshing is not eye-friendly as these sections simply blink (probably are being fully reprinted). I have noticed that when the date and time VFR opcodes are placed in a submenu (which has a title frame), the title and hotkey help sections do not blink. I have to dig a little bit more to find a way to stop refreshing of these sections, if not needed (thus draft).

SergiiDmytruk:
Maybe drop "Select Language"?  If addition of time and date doesn't break tests, this change shouldn't either and only English language is supported anyway.

miczyg1:
> Maybe drop "Select Language"? If addition of time and date doesn't break tests, this change shouldn't either and only English language is supported anyway.

French is supported to some extent though. Besides I am not sure if the language variable would be populated properly if I remove it. Also we have plans for more languages https://github.com/Dasharo/dasharo-issues/issues/526

MdeModulePkg/Application/UiApp/FrontPageStrings.uni
```diff
@@ -46,6 +48,14 @@
                                        #language fr-FR  "Choisir la Langue"
 #string STR_LANGUAGE_SELECT_HELP       #language en-US  "This is the option one adjusts to change the language for the current system"
                                        #language fr-FR  "Ceci est l'option qu'on ajuste pour changer la langue pour le système actuel"
+#string STR_DATE_PROMPT                #language en-US  "System Date"
+                                       #language fr-FR  "Date du système"
+#string STR_DATE_HELP                  #language en-US  "This is the help for the Date (month/day/year)."
+                                       #language fr-FR  "C'est l'aide pour la Date (mois/jour/année)."
+#string STR_TIME_PROMPT                #language en-US  "System Time"
+                                       #language fr-FR  "Le temps du système"
+#string STR_TIME_HELP                  #language en-US  "This is the help for the Time (hour/minute/second). Time is represented in local time zone."
```
SergiiDmytruk, line 57: Help strings for time and date need an improvement ("This is the help" part for sure, "This is the option one adjusts" for "Select Language" always looked quite ridiculus).

SergiiDmytruk (commented):
Yeah, that flickering is a bit annoying.

If we don't need an extra empty line, can make things more symmetric:

qemu

I like the changes, although would prefer more granular commits (current one is "through away stuff, bring new stuff in") or at least list of things changed in the commit message.  EDK's code is far from being readable to simply read changed lines.

Also wanted to suggest better ordering of menu items, but looks like it needs introduction of some weighs/priorities for `UiListThirdPartyDrivers()` to take into account (I thought the order is just hard-coded somewhere).

MdeModulePkg/Application/UiApp/FrontPageStrings.uni
miczyg1: Right, taken from SampleDriverDxe and didn't update it yet.

miczyg1:
> Yeah, that flickering is a bit annoying.
> 
> If we don't need an extra empty line, can make things more symmetric:
> 
> qemu
> 

Looks better indeed.

> I like the changes, although would prefer more granular commits (current one is "through away stuff, bring new stuff in") or at least list of things changed in the commit message. EDK's code is far from being readable to simply read changed lines.
> 

Yes I can do that, for now I just dumped the changes to remote and that's all.

> Also wanted to suggest better ordering of menu items, but looks like it needs introduction of some weighs/priorities for `UiListThirdPartyDrivers()` to take into account (I thought the order is just hard-coded somewhere).

It may be just an order of handles returned by HiiGetHiiHandles. I wouldn't be surprised if the order of handles is connected to the order of DXE driver execution, or even the order of the drivers inside the flash image.

macpijan:
That was some nice suggestion, perhaps we should go back to it after rebase.

PR #120, opened 2024-02-28
Vboot popup fixes

mkopec:


mkopec:
This won't work - we can't rely on the recovery flag because we're clearing it before handing off to payload. Another solution is needed.

PR #121, opened 2024-03-10
Changes for APUs

SergiiDmytruk:
APUs seem to work without these changes to EDK now, but they still might be worth keeping.  Or at least FADT change is, can drop others if they would complicate some future rebase/merge.

### DIFF
--- a/MdeModulePkg/Application/UiApp/FrontPage.c
+++ b/MdeModulePkg/Application/UiApp/FrontPage.c
@@ -528,10 +528,14 @@ UpdateFrontPageBannerStrings (
   SMBIOS_TABLE_TYPE17               *Type17Record;
   SMBIOS_TABLE_TYPE19               *Type19Record;
   EFI_SMBIOS_TABLE_HEADER           *Record;
+  UINT16                            MemorySize;
+  UINT32                            ExtendedMemorySize;
   UINT64                            InstalledMemory;
   UINT16                            MemorySpeed;
   BOOLEAN                           FoundCpu;
 
+  MemorySize = 0xFFFF; // means "unknown" in Type 17 table of SMBIOS
+  ExtendedMemorySize = 0;
   InstalledMemory = 0;
   MemorySpeed = 0;
   FoundCpu = 0;
@@ -651,6 +655,9 @@ UpdateFrontPageBannerStrings (
       if (Type17Record->ConfiguredMemoryClockSpeed > MemorySpeed) {
           MemorySpeed = Type17Record->ConfiguredMemoryClockSpeed;
       }
+
+      MemorySize = Type17Record->Size;
+      ExtendedMemorySize = Type17Record->ExtendedSize;
     }
 
     if ( Record->Type == SMBIOS_TYPE_MEMORY_ARRAY_MAPPED_ADDRESS ) {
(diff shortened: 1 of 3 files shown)


UefiPayloadPkg/AcpiPlatformDxe/AcpiPlatform.c
```diff
@@ -74,7 +74,10 @@ InstallTablesFromXsdt (
       FadtTable = (EFI_ACPI_6_3_FIXED_ACPI_DESCRIPTION_TABLE *)
                     (UINTN) CurrentTablePointer;
       DsdtTable  = (EFI_ACPI_DESCRIPTION_HEADER *) (UINTN) FadtTable->XDsdt;
+
       FacsTable = (EFI_ACPI_6_3_FIRMWARE_ACPI_CONTROL_STRUCTURE *) (UINTN) FadtTable->XFirmwareCtrl;
+      if (FacsTable == NULL)
+          FacsTable = (EFI_ACPI_6_3_FIRMWARE_ACPI_CONTROL_STRUCTURE *) (UINTN) FadtTable->FirmwareCtrl;
```
miczyg1, line 80: @mkopec does it collide somehow with these changes? https://github.com/Dasharo/edk2/pull/89/files

MdeModulePkg/Application/UiApp/FrontPage.c
```diff
@@ -667,6 +674,19 @@ UpdateFrontPageBannerStrings (
     Status = Smbios->GetNext (Smbios, &SmbiosHandle, NULL, &Record, NULL);
   }
 
+  if ( InstalledMemory == 0 && MemorySize != 0xFFFF ) {
```
miczyg1, line 677: The problem with these changes is that type17 and type19 might not always be in ascending order (hypothetically). If both types are present, we will have double the real memory capacity in the system (a sum of type17 and type19) if it is in ascending order (type17 will be counted first, then type 19 will add up more memory unconditionally). Type17 should be used as a fallback and parsed separately in a different loop or a different condition should be used to prevent such scenario.

miczyg1 (changes requested):

MdeModulePkg/Application/UiApp/FrontPage.c
SergiiDmytruk: These lines are added after the loop.  The `}` 2 lines above is for the `while` loop over ACPI tables.

miczyg1: Right, sorry @SergiiDmytruk 

It is hard to notice the scopes in the GH web GUI diff in the PR... Closing the thread.

UefiPayloadPkg/AcpiPlatformDxe/AcpiPlatform.c
miczyg1: @mkopec ping

mkopec: I don't think it collides, but it seems redundant, XFirmwareCtrl should always be non-null since https://github.com/Dasharo/coreboot/commit/a07b09ab71d04b870b9413b7e075a2369f59fd0a

mkopec: If S3 still works after these changes then it should be okay to merge. I don't think it collides, but coreboot should populate XFirmwareCtrl always anyways

mkopec (approved):

UefiPayloadPkg/AcpiPlatformDxe/AcpiPlatform.c
SergiiDmytruk: I don't think "always" applies.  See https://github.com/Dasharo/coreboot/blob/a9d4e0ab871a7c71477315829ce22076c9435e31/src/acpi/acpi.c#L1060-L1088 (top and bottom parts).  It sets them, but then calls custom functions and [the one in `hudson/`](https://github.com/Dasharo/coreboot/blob/a9d4e0ab871a7c71477315829ce22076c9435e31/src/southbridge/amd/pi/hudson/fadt.c#L55-L56) does:
```c
	fadt->x_firmware_ctl_l = 0;	/* set to 0 if firmware_ctrl is used */
	fadt->x_firmware_ctl_h = 0;
```
It works without this change probably because something else updates the fields.

miczyg1: It depends whether the FACS is above 4G or not. coreboot does not create the tables above 4G (maybe if long mode is enabled then perhaps yes). Anyways this is a good fallback if FACS 64bit address is zero, then use 32bit FACS address.

miczyg1 (approved):

PR #127, opened 2024-04-11
PlatformBootManager: handle Vboot EC errors

mkopec:


miczyg1:
LGTM, I will leave the message formula to @macpijan

mkopec:
@macpijan ping

macpijan:
How about:

```
Warning

It appears that firmware update process of the Embedded Controller has failed.
It is likely due to the AC adapter was not present. Please connect the AC
adapter, and boot the device again. 

If the AC adapter is connected, and the message persists after a few attempts,
please report this problem via one of the Dasharo support channels: https://docs.dasharo.com/#support

You may continue to use your device, although you may experience some issues
related to EC features (such as fan control, or function keys) if the EC
firmware version is not in sync with the main firmware.
```

@pietrushnic @BeataZdunczyk any thoughts?

pietrushnic:
@macpijan, do you plan to add a support section on docs.dasharo.com? I don't feel like a decision-maker regarding UX in firmware; I'm too biased, and my opinion is forming based on community and customer feedback. The message looks ok to me. It is not scary, but if it is received correctly, we can probably only verify it in the field.

BeataZdunczyk:
> How about:

LGTM @macpijan.

BeataZdunczyk (approved):

macpijan (approved):

PR #129, opened 2024-04-25
Rebase onto edk2-stable202402

SergiiDmytruk:
Not a real PR, since this is a rebase (https://github.com/Dasharo/dasharo-issues/issues/432), but could still be used for comments.

Things of note:

- [x] https://github.com/Dasharo/dasharo-issues/issues/432#issuecomment-2068222166
   - use of `BaseDebugLibSerialPort` now causes build failures due to build cycles and its use in one case was meant to fix an issue with QubesOS
   - applies to https://github.com/Dasharo/edk2/pull/129/commits/0eb481acee2f47f516372c7cc3f7fbb17fc8e019 as well
   - solution:
      * ~~revert that commit: https://github.com/Dasharo/edk2/pull/129/commits/4df31e548c053646c07633248e92494ae046dc53~~ (technically it is reverted)
      * change 4df31e548c053646c07633248e92494ae046dc53 to have the same effect but still use `PeiDxeDebugLibReportStatusCode` by making it do nothing in runtime phase https://github.com/Dasharo/edk2/pull/129/commits/63900287f1d0e4029f899d27372c29cb81446019
      * use PeiDxeDebugLibReportStatusCode library as DebugLib where use of BaseDebugLibSerialPort causes 
dependency cycle: https://github.com/Dasharo/edk2/pull/129/commits/d13b2f3387af324fcdd8c1047495dc43ad5689ac
- [ ] In light of Dasharo for Q35, possibly drop all OVMF changes
- [x] https://github.com/Dasharo/dasharo-issues/issues/432#issuecomment-2078209639
   - OpenSSL3 is significantly larger than OpenSSL1, but there is MbedTLS now
   - solution: switched to MbedTLS (e.g., [b634e05](https://github.com/Dasharo/edk2/pull/129/commits/b634e05db985adc07c3ded21c96279c0d0ad5845), but mind rebases)
- [x] https://github.com/Dasharo/dasharo-issues/issues/432#issuecomment-2089199396
   - MbedTLS might be missing stuff
   - solution: pull recent upstream commits with add missing parts https://github.com/Dasharo/dasharo-issues/issues/432#issuecomment-2143058359
- [x] `PcdResetOnMemoryTypeInformationChange` needs to be disabled for boot to work
   - disabled in https://github.com/Dasharo/edk2/pull/129/commits/765c152b0af04084e207fe90f4df902a3425cb79
- [x] MbedTLS uses BootServices after they are gone https://github.com/Dasharo/edk2/pull/129#issuecomment-2153493525
    - fixed https://github.com/Dasharo/edk2/pull/129#issuecomment-2156200987

Issues discovered while testing that must be resolved:
- [x] `HobList != ((void *) 0)` assertions in coreboot+OVMF and at least on APU6 (MSI and NovaCustom V540TU somehow worked). https://github.com/Dasharo/edk2/pull/129#issuecomment-2135503704
   - workaround: c10b9731663e7b31161a1e5619d4894815df26e3
- [x] Builtin PS/2 keyboard isn't working in UEFI https://github.com/Dasharo/edk2/pull/129#issuecomment-2135111897
   - likely fixed https://github.com/Dasharo/edk2/pull/129#issuecomment-2144031287
- [x] RAM size is zero at least on NovaCustom V540TU. https://github.com/Dasharo/edk2/pull/129#issuecomment-2135111897
   - doesn't seem to be related to the rebase https://github.com/Dasharo/edk2/pull/129#issuecomment-2147256194
- [x] UEFI BootManagerMenuApp entry visible in one time boot menu.  https://github.com/Dasharo/edk2/pull/129#issuecomment-2135111897
   - fixed https://github.com/Dasharo/edk2/pull/129#issuecomment-2156800176
- [x] No TPM menu.  https://github.com/Dasharo/edk2/pull/129#issuecomment-2135111897
   - likely fixed https://github.com/Dasharo/edk2/pull/129#issuecomment-2144031287
- [x] Difficulties with iPXE: there is no way to get to its shell https://github.com/Dasharo/edk2/pull/129#issuecomment-2176471325
    - fixed https://github.com/Dasharo/edk2/pull/129#issuecomment-2179212147
- [x] It seems that after enabling SB it gets back to disabled state after a reboot. https://github.com/Dasharo/edk2/pull/129#issuecomment-2176471325
    - fixed https://github.com/Dasharo/edk2/pull/129#issuecomment-2179212147
- [x] (possible) No pre-installed boot options?  It might be OK if it's not ESP or a media.


mkopec:
Looks pretty good already, boots on NovaCustom V540TU (with https://github.com/Dasharo/edk2/pull/136)

Some issues I've noticed already:
- RAM capacity on the frontpage is 0
- UEFI BootManagerMenuApp entry visible in one time boot menu
- Builtin PS/2 keyboard isn't working in UEFI
  - Works after manually enabling it for ConIn in BootMaintenanceManager
- No TPM menu

krystian-hebel:
It doesn't work on coreboot with edk2 as a payload on QEMU, failing very early:

```
[INFO ]  CBFS: Found 'fallback/payload' @0x80 size 0x188061 in mcache @0x7fefe9ac
[DEBUG]  Checking segment from ROM address 0xff8002ac
[DEBUG]  Checking segment from ROM address 0xff8002c8
[DEBUG]  Loading segment from ROM address 0xff8002ac
[DEBUG]    code (compression=1)
[DEBUG]    New segment dstaddr 0x00800000 memsize 0xe00000 srcaddr 0xff8002e4 filesize 0x188029
[DEBUG]  Loading Segment: addr: 0x00800000 memsz: 0x0000000000e00000 filesz: 0x0000000000188029
[DEBUG]  using LZMA
[DEBUG]  Loading segment from ROM address 0xff8002c8
[DEBUG]    Entry Point 0x00800850
[DEBUG]  BS: BS_PAYLOAD_LOAD run times (exec / console): 142 / 4 ms
[DEBUG]  ICH-NM10-PCH: watchdog disabled
[DEBUG]  Jumping to boot code at 0x00800850(0x7fe86000)
ASSERT [PeiCore] /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Library/PeiHobLib/HobLib.c(45): HobList != ((void *) 0)
ASSERT [PeiCore] /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Library/PeiHobLib/HobLib.c(76): HobStart != ((void *) 0)
```

@SergiiDmytruk any idea what may be causing it? I haven't seen too much changes in such early code done by us, so maybe this is something that changed upstream.

On a tangent note, there is `OvmfPkg/Library/PeilessStartupLib`, perhaps such approach makes sense for our use case as well?

SergiiDmytruk:
@mkopec, thanks, will take a look whether these are caused by a missing commit, conflict resolution gone wrong or new base needs somewhat different changes.

> @SergiiDmytruk any idea what may be causing it? I haven't seen too much changes in such early code done by us, so maybe this is something that changed upstream.

Not really, seems like upstream changes.  Weird thing is that I just tried using rebased EDK on APU6 to check whether RAM size is reported and saw exactly the same asserts.  I'd expect this to either work or not work for `DasharoPayloadPkg`.

> On a tangent note, there is OvmfPkg/Library/PeilessStartupLib, perhaps such approach makes sense for our use case as well?

Don't we need PEI for TPM drivers of EDK?  IIUC UefiPayloadPkg in upstream now has no PEI which is why it can't be used by us without reworking code somehow.

SergiiDmytruk:
> Not really, seems like upstream changes.

Well, maybe indirectly...  The code is old, but something must be different.  The error occurs at https://github.com/Dasharo/edk2/blob/9829f826a3542c88ce1b85ea03443f268717be90/DasharoPayloadPkg/Library/AcpiTimerLib/AcpiTimerLib.c#L44

SergiiDmytruk:
It shouldn't fail there for OVMF which uses its own `AcpiTimerLib` without accessing HOBs, but something similar is probably happening.  On APU `BlSupportPei` seems to not be executed as if dependencies of PEIMs are not resolved correctly.

SergiiDmytruk:
So `AcpiTimerLibConstructor()` needs HOBs which are created by `BlSupportPei` which is loaded by PEI core, but in the rebased EDK `AcpiTimerLib` is somehow necessary for PEI core to even start its execution (not sure what causes this dependency).

After working around that APU is able to boot, although it seems to have issues with input/output (no boot menu, can't enter setup UI).  Q35 seems to act similar to APU.  Should probably find out what caused this new dependency in the first place to find a proper mitigation.

SergiiDmytruk:
> Should probably find out what caused this new dependency in the first place to find a proper mitigation.

There is probably no specific reason.  Older version happened to work without trouble in the configuration used by coreboot and the new one just doesn't because code has changed.  The workaround in 5ffa33a4f5 is probably the way to go.

SergiiDmytruk:
Looks like issue with serial on APU was really an issue with serial consoles in rebased version, I just used KVM and didn't notice it initially.  Turned out that 35e5619223 doesn't actually revert 4628519c55.  The latter moves `PlatformConsoleInit()` from top to bottom while the former moves it just couple statements up.  The change should be present for serial consoles to be registered properly and now serial console gets enabled automatically.

@krystian-hebel, Q35 seems to work now.

@mkopec, I see RAM size on MSI Z690-A DDR5, Q35 and APU4.  Was it displayed before the rebase?

> * Builtin PS/2 keyboard isn't working in UEFI
>   * Works after manually enabling it for ConIn in BootMaintenanceManager

I think this could have been fixed along with serial output, PS/2 keyboard seems to be registered in `PlatformConsole.c` as well and it was invoked before devices were discovered.

> No TPM menu

No "Device Manager" -> "TCG2 Configuration" menu?  Was present on MSI but not on APU4 due to TPM error:
```
TPM error!
Error: Image at 0007E219000 start failed: Device Error
```
Turned out that f6ba45d6c9 commit was necessary, at least APU now detects TPM and probably NovaCustom V540TU.

> UEFI BootManagerMenuApp entry visible in one time boot menu

Still couldn't find what was suppressing it from appearing on "One time boot" menu.

krystian-hebel:
> So `AcpiTimerLibConstructor()` needs HOBs which are created by `BlSupportPei` which is loaded by PEI core, but in the rebased EDK `AcpiTimerLib` is somehow necessary for PEI core to even start its execution (not sure what causes this dependency).

Is it possible to add `BlSupportPei` to `APRIORI PEI`, together with PCD as it is its dependency?

SergiiDmytruk:
> Is it possible to add `BlSupportPei` to `APRIORI PEI`, together with PCD as it is its dependency?

Doesn't work when I try:
1. coreboot requires `SecCore` to be the first one in UEFI payload:
   * `E: Not a usable UEFI firmware volume.` on build
   * `[EMERG]  Payload not loaded` when run
2. Making `APRIORI PEI {` follow `SecCore` in FDF file breaks EDK autogen probably because `APRIORI PEI` must be the first one if present.

Not sure how hard it would be to adjust coreboot to be more flexible, it must have trouble finding entrypoint.  And in the end it might not help, because APRIORI PEI changes ordering in which PEIMs get processed and those asserts followed by segfault happen before PEIMs dispatcher even starts.  The timer calls probably happen due to performance measurement code somewhere running as part of PEI core.

krystian-hebel:
Few remaining asserts can be removed by dropping `PerformanceLib` from PEI_CORE, I believe the same problem occurs in SEC, except it isn't printed due to `BaseDebugLibNull`.

It still fails for me on Q35, although much further:

```
Loading driver at 0x000069F1000 EntryPoint=0x000069FBA31 dpDynamicCommand.efi
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF 6BB9B98
InstallProtocolInterface: 6A1EE763-D47A-43B4-AABE-EF1DE2AB56FC 69FFCB0
ProtectUefiImageCommon - 0x6BB8B40
  - 0x00000000069F1000 - 0x00000000000111C0
PROGRESS CODE: V03040002 I0
InstallProtocolInterface: 3C7200E9-005F-4EA4-87DE-A3DFAC8A27C3 69FF2C0
PROGRESS CODE: V03040003 I0
!!!! X64 Exception Type - 0E(#PF - Page-Fault)  CPU Apic ID - 00000000 !!!!
ExceptionData - 0000000000000010  I:1 R:0 U:0 W:0 P:0 PK:0 SS:0 SGX:0
RIP  - FFFFFFFFC7CCD4A0, CS  - 0000000000000038, RFLAGS - 0000000000000286
RAX  - FFFFFFFFC7CCD4A0, RCX - 0000000000000080, RDX - 0000000007C9FC30
RBX  - 000000000746EA98, RSP - 0000000007C9FC20, RBP - 0000000007C9FD70
RSI  - 00000000070A6C18, RDI - 0000000000000001
R8   - 0000000007C9FC80, R9  - 000000000746E928, R10 - 00000000000003F8
R11  - 0000000000000010, R12 - 000000000746E929, R13 - 0000000000000000
R14  - 000000000746EAA0, R15 - 0000000007CCD4A0
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010011, CR2 - FFFFFFFFC7CCD4A0, CR3 - 0000000007601000
CR4  - 0000000000000228, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 00000000000F0400
GDTR - 00000000075EC000 0000000000000047, LDTR - 0000000000000000
IDTR - 000000000709C018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 0000000007C9F880
!!!! Find image based on IP(0x6BB8E18) (No PDB)  (ImageBase=0000000006BAD018, EntryPoint=0000000006BB47C5) !!!!
```

After removing `dpDynamicCommand.efi` it failed in the same way on `tftpDynamicCommand.efi`, and after removing that:

```
Loading driver at 0x00006A1B000 EntryPoint=0x00006A209A4 OpalPasswordDxe.efi
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF 7039E18
ProtectUefiImageCommon - 0x7039AC0
  - 0x0000000006A1B000 - 0x0000000000018500
PROGRESS CODE: V03040002 I0
InstallProtocolInterface: 18A031AB-B443-4D1A-A5C0-0C09261E9F71 6A33100
InstallProtocolInterface: 107A772C-D5E1-11D4-9A46-0090273FC14D 6A32FB0
InstallProtocolInterface: 6A7A5CFF-E8D9-4F70-BADA-75AB3025CE14 6A32F90
InstallProtocolInterface: 330D4706-F2A0-4E4F-A369-B66FA8D54385 6A331E0
InstallProtocolInterface: 09576E91-6D3F-11D2-8E39-00A0C969723B 6A32FE0
PROGRESS CODE: V03040003 I0
!!!! X64 Exception Type - 0E(#PF - Page-Fault)  CPU Apic ID - 00000000 !!!!
ExceptionData - 0000000000000010  I:1 R:0 U:0 W:0 P:0 PK:0 SS:0 SGX:0
RIP  - FFFFFFFFC7CCD4A0, CS  - 0000000000000038, RFLAGS - 0000000000000286
RAX  - FFFFFFFFC7CCD4A0, RCX - 0000000000000080, RDX - 0000000007C9FC30
RBX  - 000000000746E098, RSP - 0000000007C9FC20, RBP - 0000000007C9FD70
RSI  - 00000000070DB298, RDI - 0000000000000001
R8   - 0000000007C9FC80, R9  - 000000000746E8A8, R10 - 00000000000003F8
R11  - 0000000000000010, R12 - 000000000746E8A9, R13 - 0000000000000000
R14  - 000000000746E0A0, R15 - 0000000007CCD4A0
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010011, CR2 - FFFFFFFFC7CCD4A0, CR3 - 0000000007601000
CR4  - 0000000000000228, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 00000000000F0400
GDTR - 00000000075EC000 0000000000000047, LDTR - 0000000000000000
IDTR - 00000000070BD018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 0000000007C9F880
!!!! Find image based on IP(0x7039D18) /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/DEBUG_COREBOOT/X64/MdeModulePkg/Universal/FaultTolerantWriteDxe/FaultTolerantWriteDxe/DEBUG/FaultTolerantWriteDxe.dll (ImageBase=0000000007033000, EntryPoint=0000000007036F9A) !!!!
```

I'm not sure how image lookup works, but according to logs 0x7039D18 isn't `FaultTolerantWriteDxe` code, it is `.debug_info` in `.debug` file which is stripped from `.efi`. Nevertheless, after stopping execution and analyzing memory, 0x7039D18 has ASCII 'hndl' followed by mostly zeros, it isn't valid x86 code.

`gEfiHiiDatabaseProtocolGuid` is visible on stack, along with pointers to addresses that don't seem to belong any driver whose base address has been logged. Most of those seem to be magic sequences (`hndl`, `evnt`, `ptal`). Last address that I'm 100% sure is code is instruction immediately following call to `CoreDispatcher` done by `DxeCore.efi`, and I'm pretty sure that the page fault happens somewhere in that function.

krystian-hebel:
Interestingly, after changing the amount of RAM I got different panic:

```
!!!! X64 Exception Type - 06(#UD - Invalid Opcode)  CPU Apic ID - 00000000 !!!!
RIP  - 000000007BD4857D, CS  - 0000000000000038, RFLAGS - 0000000000010282
RAX  - 000000007F46E098, RCX - 00000000000000E0, RDX - 000000007FC9FC30
RBX  - 000000003FCCD494, RSP - 000000007FC9FC20, RBP - 000000007FC9FD70
RSI  - 000000007F0DB298, RDI - 0000000000000001
R8   - 000000007FC9FC80, R9  - 000000007F46E8A8, R10 - 00000000000003F8
R11  - 0000000000000010, R12 - 000000007F46E8A9, R13 - 0000000000000000
R14  - 000000007F46E0A0, R15 - 000000007FCCD4A0
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010011, CR2 - 0000000000000000, CR3 - 000000007F601000
CR4  - 0000000000000228, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 00000000000F0400
GDTR - 000000007F5EC000 0000000000000047, LDTR - 0000000000000000
IDTR - 000000007F0BD018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 000000007FC9F880
!!!! Find image based on IP(0x7BD4857D) /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/DEBUG_COREBOOT/X64/CrScreenshotDxe/CrScreenshotDxe/DEBUG/CrScreenshotDxe.dll (ImageBase=0000000000E50498, EntryPoint=0000000000E57E99) !!!!
```

SergiiDmytruk:
> Few remaining asserts can be removed by dropping `PerformanceLib` from PEI_CORE, I believe the same problem occurs in SEC, except it isn't printed due to `BaseDebugLibNull`.

Well, I assumed it was added for a purpose hence the workaround instead of removing performance measurements.
 
> It still fails for me on Q35, although much further:

Weird given that it works for me on your `fw_update_smi2` branch with https://github.com/Dasharo/coreboot/commit/455debb9384ad2e17b587f49fc3aca1d9c562764 cherry-picked on top of it and a tiny change, see the spoiler below.

<details><summary>Kconfig change, defconfig, bash script</summary>
<p>

One more change is
```diff
diff --git a/src/vendorcode/intel/Kconfig b/src/vendorcode/intel/Kconfig
index b481aa0090b..f09aebea2cd 100644
--- a/src/vendorcode/intel/Kconfig
+++ b/src/vendorcode/intel/Kconfig
@@ -8,6 +8,7 @@ config UEFI_2_4_BINDING
 	select UDK_BASE
 
 config UDK_2017_BINDING
+	def_bool y if BOARD_EMULATION_QEMU_X86_Q35
 	def_bool n
 	select UDK_BASE
 
```

defconfig:
```
CONFIG_OPTION_BACKEND_NONE=y
CONFIG_BOARD_EMULATION_QEMU_X86_Q35=y
# CONFIG_DASHARO_FIRMWARE_UPDATE_MODE is not set
CONFIG_VGA_TEXT_FRAMEBUFFER=y
CONFIG_RESOURCE_ALLOCATION_TOP_DOWN=y
CONFIG_DRIVERS_EFI_VARIABLE_STORE=y
CONFIG_PAYLOAD_EDK2=y
CONFIG_EDK2_REPOSITORY="https://github.com/Dasharo/edk2.git"
CONFIG_EDK2_TAG_OR_REV="d471c3f6626ec99dfda3993da8ea53f5bf2c291d"
CONFIG_EDK2_USE_EDK2_PLATFORMS=y
CONFIG_EDK2_PLATFORMS_REPOSITORY="https://github.com/Dasharo/edk2-platforms"
CONFIG_EDK2_PLATFORMS_TAG_OR_REV="3323ed481d35096fb6a7eae7b49f35eff00f86cf"
CONFIG_EDK2_DEBUG=y
# CONFIG_EDK2_SECURE_BOOT_DEFAULT_ENABLE is not set
CONFIG_EDK2_DASHARO_SYSTEM_FEATURES=y
CONFIG_EDK2_DASHARO_SERIAL_REDIRECTION_DEFAULT_ENABLE=y
```

Script:
```bash
#!/bin/bash
params=(
    -M q35,smm=on,max-fw-size=16M
    -enable-kvm
    -m 4096
    -drive if=pflash,format=raw,unit=0,file=build/coreboot.rom
    -serial stdio
    -action reboot=shutdown
)

exec qemu-system-x86_64 "${params[@]}"
```

</p>
</details>

mkopec:
TPM and PS/2 KB issues are fixed, thanks.

> @mkopec, I see RAM size on MSI Z690-A DDR5, Q35 and APU4. Was it displayed before the rebase?

I saw it on a different device with different DIMM modules, but when testing on the same unit, it's not displayed on old and new edk2. It's probably an issue with MTL and DDR5, not with EDK2

krystian-hebel:
Interesting, it seems to work with `coreboot-sdk:2022-09-18_c8870b1334` but not with `coreboot-sdk:2023-11-24_2731fa619b`. `2021-09-23_b0d87f753c` fails due to the lack of imagemagick, and `2024-05-20_b4949d3de5` can't build iPXE, but image built without it also works. I haven't tried other versions.

---

`2023-11-24_2731fa619b` uses `gcc version 13.2.0 (Debian 13.2.0-7)`, both `2022-09-18_c8870b1334` (older) and `2024-05-20_b4949d3de5` (newer) use 12.2.0 (`Debian 12.2.0-3` and `Debian 12.2.0-14` respectively). Perhaps whatever was hidden by any of `-Wno-vla-parameter -Wno-stringop-overflow -Wno-use-after-free -Wno-dangling-pointer` is UB or plain bugs that should be fixed.

krystian-hebel:
One more problem on QEMU, which may be connected to a problem that @miczyg1 was debugging on Protectli.
Zrzut ekranu z 2024-06-05 15-24-54
Zrzut ekranu z 2024-06-05 15-24-59
Zrzut ekranu z 2024-06-05 15-25-07
Zrzut ekranu z 2024-06-05 15-25-12

By checking CR2 address against log of loaded EFI modules, I found out that this is `DxeCore`, in particular `CoreAllocatePool()`:

```
DXE IPL Entry
Loading PEIM D6A2CB7F-6A18-4E2F-B43B-9920A733700A
Loading PEIM at 0x0007FC9E000 EntryPoint=0x0007FCBAE5B DxeCore.efi
PROGRESS CODE: V03021001 I0
Loading DXE CORE at 0x0007FC9E000 EntryPoint=0x0007FCBAE5B
AddressBits=40 5LevelPaging=0 1GPage=0
Pml5=1 Pml4=2 Pdp=512 TotalPage=1027
Install PPI: 605EA650-C65C-42E1-BA80-91A52AB618C6
Notify: PPI Guid: 605EA650-C65C-42E1-BA80-91A52AB618C6, Peim notify entry point: 810B84
HandOffToDxeCore() Stack Base: 0x7FC7E000, Stack Size: 0x20000
PROGRESS CODE: V03040003 I0
```
Zrzut ekranu z 2024-06-05 16-03-15

By comparing code from Linux's panic (in particular address modulo page size) it seems to be called from `VariableRuntimeDxe`:

```
0000000000003ada <InternalAllocatePool>:
    3ada:       55                      push   %rbp
    3adb:       89 f9                   mov    %edi,%ecx
    3add:       48 89 f2                mov    %rsi,%rdx
    3ae0:       48 89 e5                mov    %rsp,%rbp
    3ae3:       48 83 ec 30             sub    $0x30,%rsp
    3ae7:       48 8b 05 32 da 01 00    mov    0x1da32(%rip),%rax        # 21520 <gBS>
    3aee:       4c 8d 45 f8             lea    -0x8(%rbp),%r8
    3af2:       ff 50 40                callq  *0x40(%rax)
    3af5:       48 83 c4 20             add    $0x20,%rsp
    3af9:       48 85 c0                test   %rax,%rax
    3afc:       79 08                   jns    3b06 <InternalAllocatePool+0x2c>
    3afe:       48 c7 45 f8 00 00 00    movq   $0x0,-0x8(%rbp)
    3b05:       00 
    3b06:       48 8b 45 f8             mov    -0x8(%rbp),%rax
    3b0a:       c9                      leaveq 
    3b0b:       c3                      retq
```

This code tries to use Boot Services after call to `ExitBootServices()`, which obviously shouldn't happen. Wrong or bugged `MemoryAllocationLib` maybe?

I haven't checked whether this is something limited to Dasharo or if it applies to upstream edk2 as well.

krystian-hebel:
Ubuntu 24.04 works slightly better (above was on 22.04), it can read UEFI variables, but not write to them nor create new ones. There is no oops in dmesg. I think it may use static copy of variables obtained on boot, or something else changed in the way efivars are used in newer kernel.

SergiiDmytruk:
> Perhaps whatever was hidden by any of `-Wno-vla-parameter -Wno-stringop-overflow -Wno-use-after-free -Wno-dangling-pointer` is UB or plain bugs that should be fixed.

Could be.  I build with my system's 14.1.0.

> This code tries to use Boot Services after call to `ExitBootServices()`, which obviously shouldn't happen. Wrong or bugged `MemoryAllocationLib` maybe?

https://github.com/tianocore/edk2/blob/90cb1ec33225a070e9fea1d94c72ff590bd38731/MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf#L21:
```
LIBRARY_CLASS                  = MemoryAllocationLib|DXE_DRIVER DXE_RUNTIME_DRIVER DXE_SMM_DRIVER UEFI_APPLICATION UEFI_DRIVER
```

https://github.com/tianocore/edk2/blob/90cb1ec33225a070e9fea1d94c72ff590bd38731/MdePkg/Library/UefiMemoryAllocationLib/MemoryAllocationLib.c#L43
```c
  Status = gBS->AllocatePages (AllocateAnyPages, MemoryType, Pages, &Memory);
```

I'm not sure if page allocation should be done by runtime services though, maybe the invoking code is wrong.  Also I don't see a more suitable memory library.

krystian-hebel:
I'm getting the same behavior on MTL-H laptop, with expected deviations caused by different memory layout. It worked fine with older edk2 before rebase.

I've looked at the recent changes done to RuntimeDxe and didn't see anything obviously wrong with them. It may indeed be the calling code. Some of those allocations are to be done once, on first call, maybe this is what's missing. On the other hand, this tries to allocate BootServicesData... Maybe it was always broken, but worked by accident because allocated pool was on the same page as RuntimeServicesData, but IIRC UEFI allocates whole pages event for AllocatePool.

There is also `AtRuntime()` that controls execution of some functions, it is another possible cause.

SergiiDmytruk:
Was very puzzled on seeing the crash even after adding `if (EfiAtRuntime()) return EFI_INVALID_PARAMETER;` to about every function assigned to `gRT` and turned out that `VariableDxe.c` uses `SystemTable->RuntimeServices` to set them up instead...  Did get a backtrace in the end pointing at MbedTLS as an offender (frames 3 and 4):

```
#0  0x000000007f8fbaf2 in InternalAllocatePool (MemoryType=MemoryType@entry=EfiBootServicesData, AllocationSize=AllocationSize@entry=3340)
    at $EDK_PATH/MdePkg/Library/UefiMemoryAllocationLib/MemoryAllocationLib.c:376
#1  0x000000007f8fbba4 in InternalAllocateZeroPool (PoolType=PoolType@entry=EfiBootServicesData, AllocationSize=AllocationSize@entry=3340)
    at $EDK_PATH/MdePkg/Library/UefiMemoryAllocationLib/MemoryAllocationLib.c:469
#2  0x000000007f8fbc0e in AllocateZeroPool (AllocationSize=3340) at $EDK_PATH/MdePkg/Library/UefiMemoryAllocationLib/MemoryAllocationLib.c:496
#3  0x000000007f90cee8 in WrapPkcs7Data (P7Data=0xffff94690aaa0028 "0\202\f\365\002\001\0011\0170\r\006\t`\206H\001e\003\004\002\001\005", P7Length=3321, WrapFlag=0xffffac7e4006b32f ";\f\r", WrapData=0xffffac7e4006b338, WrapDataSize=0xffffac7e4006b330)
    at $EDK_PATH/CryptoPkg/Library/BaseCryptLibMbedTls/Pk/CryptPkcs7VerifyCommon.c:992
#4  0x000000007f90d66b in Pkcs7GetSigners (P7Data=P7Data@entry=0xffff94690aaa0028 "0\202\f\365\002\001\0011\0170\r\006\t`\206H\001e\003\004\002\001\005", P7Length=P7Length@entry=3321, CertStack=CertStack@entry=0xffffac7e4006ba70, 
    StackLength=StackLength@entry=0xffffac7e4006ba78, TrustedCert=0xffffac7e4006ba60, CertLength=0xffffac7e4006ba68)
    at $EDK_PATH/CryptoPkg/Library/BaseCryptLibMbedTls/Pk/CryptPkcs7VerifyCommon.c:1211
#5  0x000000007f9102d9 in Pkcs7GetSigners (CertLength=0xffffac7e4006ba68, TrustedCert=0xffffac7e4006ba60, StackLength=0xffffac7e4006ba78, CertStack=0xffffac7e4006ba70, P7Length=3321, 
    P7Data=0xffff94690aaa0028 "0\202\f\365\002\001\0011\0170\r\006\t`\206H\001e\003\004\002\001\005") at $EDK_PATH/SecurityPkg/Library/AuthVariableLib/AuthService.c:2199
#6  VerifyTimeBasedPayload (VarPayloadSize=<synthetic pointer>, VarPayloadPtr=<synthetic pointer>, OrgTimeStamp=0x7f94096c, AuthVarType=AuthVarTypePk, Attributes=103, DataSize=<optimized out>, Data=<optimized out>, VendorGuid=0xffff94690006b400, 
    VariableName=0xffff94690006b000) at $EDK_PATH/SecurityPkg/Library/AuthVariableLib/AuthService.c:2199
#7  VerifyTimeBasedPayloadAndUpdate (VariableName=0xffff94690006b000, VendorGuid=0xffff94690006b400, Data=Data@entry=0xffff94690aaa0000, DataSize=<optimized out>, Attributes=103, AuthVarType=AuthVarTypePk, VarDel=0xffffac7e4006bbcf "")
    at $EDK_PATH/SecurityPkg/Library/AuthVariableLib/AuthService.c:2487
#8  0x000000007f910c23 in ProcessVarWithPk (VariableName=0xffff94690006b000, VendorGuid=0xffff94690006b400, Data=0xffff94690aaa0000, DataSize=7085, Attributes=103, IsPk=<optimized out>)
    at $EDK_PATH/SecurityPkg/Library/AuthVariableLib/AuthService.c:768
#9  0x000000007f8fef92 in ProcessVarWithPk (IsPk=0 '\000', Attributes=103, DataSize=7085, Data=0xffff94690aaa0000, VendorGuid=0xffff94690006b400, VariableName=0xffff94690006b000)
    at $EDK_PATH/SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c:435
#10 AuthVariableLibProcessVariable (Attributes=103, DataSize=7085, Data=0xffff94690aaa0000, VendorGuid=0xffff94690006b400, VariableName=0xffff94690006b000)
    at $EDK_PATH/SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c:439
#11 VariableServiceSetVariable (VariableName=0xffff94690006b000, VendorGuid=0xffff94690006b400, Attributes=103, DataSize=7085, Data=0xffff94690aaa0000)
    at $EDK_PATH/MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c:2896
```

The failure occurs when Ubuntu updates `dbx` which requires processing of authenticated variables.

OpenSSL version uses `malloc()` in this case which is unavailable for MbedTLS.  OpenSSL's implementation is [here](https://github.com/Dasharo/edk2/blob/176c51dc69645857ed51f76ba9ac67b55c45c18c/CryptoPkg/Library/BaseCryptLib/SysCall/RuntimeMemAllocation.c#L393) and we probably need to reuse that file for MbedTLS and update other allocations as well.

SergiiDmytruk:
> OpenSSL version uses `malloc()` in this case which is unavailable for MbedTLS.

Actually, a copy of it exists and is used for MbedTLS as well, but it wasn't linked to PKCS7 implementation.  This was easy to fix, however code of MbedTLS also invokes `AllocateZeroPool()` so now it fails in the same place but with a different stack.  

Updating that to unconditional `malloc()` breaks boot, so will try to extract one file to 2 libraries to be able to use a different implementation for `DXE_RUNTIME_DRIVER` which seems to be the only way to do this because MbedTLS is of `BASE` type.

SergiiDmytruk:
That worked (no more crashes in `dmesg`) but SecureBoot got broken because compiler optimizes away `malloc()` in PKCS7 implementation thinking that it can't work.  Will need to make use of `malloc()` only in runtime phase in that file as well, not sure yet how.

SergiiDmytruk:
> Will need to make use of `malloc()` only in runtime phase in that file as well, not sure yet how.

Reused the same libraries.  There are no more page faults, but EFI variables aren't writable as in 24.04 case.

The reason is that SMI handler doesn't respond and EDK produces `EFI_NO_RESPONSE` which somehow ends up being reported as invalid parameter on command-line.

SMI handler doesn't work in runtime phase because Linux executes `SetVirtualAddressMap()` which relocates EFI runtime by moving it to `0xFFFFFFFE80000000` that a 32-bit SMI handler stub (https://github.com/Dasharo/coreboot/blob/a76d969c0717381e805515ea0e009ad58252bb80/src/cpu/x86/smm/smm_stub.S#L211-L216) just can't handle: upper half is lost and SMI gets ignored.

I tried to expose two tables from EDK to tell Linux to not relocate EFI runtime and deprecated `EFI_PROPERTIES_TABLE` did achieve that, as did passing `efi=novamap` kernel parameter from GRUB.  This however didn't make anything work, I got `CPA detected W^X violation` in `dmesg` (seems to happen on doing one-to-one mapping) and a non-working EFI vars again.

So I attempted to make coreboot run in long mode by setting `ARCH_ALL_STAGES_X86_64`, which resulted in a reboot on SMI after `SetVirtualAddressMap()`.  @krystian-hebel, should https://github.com/Dasharo/coreboot/pull/514 work fine in this case?

By the way, I know that it's virtual addressing because if I break `ConvertPointer()` EDK keeps using original addresses and everything works (maybe messing up some memory in the process).

To sum up:
 * page faults related to MbedTLS were fixed
 * 32-bit coreboot's SMI handler doesn't work after `SetVirtualAddressMap()`
 * Linux doesn't work without `SetVirtualAddressMap()`
 * 64-bit coreboot's SMI handler doesn't work after `SetVirtualAddressMap()` either
 * it doesn't seem to be EDK's fault

SergiiDmytruk:
> UEFI BootManagerMenuApp entry visible in one time boot menu

Finally found that it was caused by 4ef72fe0222fe2333c6c2249ed614310fd17c193.  And updated "DasharoPayloadPkg: boot manager improvements" (9a38a59654cf750067f1ba308edfe70ad02a3902) and one commit after that to use `LOAD_OPTION_HIDDEN`.  Also simplified that code somewhat by removing parameters to functions that only do one thing.

Also dropped one useless commit and squashed a couple.

krystian-hebel:
> So I attempted to make coreboot run in long mode by setting `ARCH_ALL_STAGES_X86_64`, which resulted in a reboot on SMI after `SetVirtualAddressMap()`. @krystian-hebel, should [Dasharo/coreboot#514](https://github.com/Dasharo/coreboot/pull/514) work fine in this case?

There were some TODs/FIXMEs related to long mode in SMM, mostly related to size requirements for page tables. Using `ARCH_ALL_STAGES_X86_64` is a big hammer, this may break more unrelated things than it fixes (FSP, payload handoff etc.) in a way that isn't immediately visible. Having only SMM running in long mode may be safer.

That said, virtual addresses in SMM are different than virtual addresses in UEFI driver. SMM should always operate on physical address, at least in the current form of handler. AFAICT this is what should be happening in https://github.com/Dasharo/edk2/pull/129/files#diff-086d49f388810e96c8b73f4786475f8c1e05ec37ab93ec47a1b09db91d6abfb3R182. Perhaps https://github.com/Dasharo/edk2/pull/129/files#diff-086d49f388810e96c8b73f4786475f8c1e05ec37ab93ec47a1b09db91d6abfb3R275 should be removed, I think that at this point it isn't consumed by UEFI. If it is, perhaps a second field with physical address should be added, or if it breaks compatibility, UEFI driver should store virtual address of `ComBuffer` locally in private variable.

SergiiDmytruk:
> Having only SMM running in long mode may be safer.

IIUC SMM code is compiled as RAM code, so you can't single it out like that and looking at `Kconfig` I wasn't sure mixing different bitness is a good idea.  Tried making only ramstage 64-bit and Qemu didn't boot.

> SMM should always operate on physical address, at least in the current form of handler.

Thanks, I missed that it used separate variable in the call.

> Perhaps https://github.com/Dasharo/edk2/pull/129/files#diff-086d49f388810e96c8b73f4786475f8c1e05ec37ab93ec47a1b09db91d6abfb3R275 should be removed, I think that at this point it isn't consumed by UEFI. If it is, perhaps a second field with physical address should be added, or if it breaks compatibility, UEFI driver should store virtual address of ComBuffer locally in private variable.

`mSmmStoreInfo->ComBuffer` is used in `SmmStoreLib{Read,Write}()` and it's not passed to SMI handler.  The whole `mSmmStoreInfo` is private data of the module.

_It's 64-bit instruction address which causes search in `amd64_apmc_node()` to fail._  Disabling pointer conversion had effect on data as well, but it's pointers to functions which matter and determine the result.  If I replace `continue` in the handler with `return node;` things work without modifications to EDK.

krystian-hebel:
Regarding MbedTLS vs OpenSSL, do we have to be so concerned with increased size? Wouldn't it be enough to make FVs bigger in FDF file to make everything fit?

SergiiDmytruk:
The problem is not with FDF, FVs don't get full, CBFS just doesn't have enough space for LZMA-compressed EDK with OpenSSL3.

krystian-hebel (approved):
I've retested QEMU on newest code ~and issue with UEFI variables indeed seems to be fixed~. If nobody sees any other problems, I think this is ready to be "merged".

EDIT: or maybe not. There are no traces of error in dmesg, but GRUB installation still fails :/

SergiiDmytruk:
EFI vars can't work in QEMU because SMI handler fails to find CPU node which send request when the calling code is above 4GiB boundary.  As a workaround, maybe could return node `0` if there is only one node.

SergiiDmytruk:
Solved revert of Xen bugfix by doing it differently: teaching `RuntimeDxeReportStatusCodeLib` to not send data to potential coreboot console in runtime phase.   `RuntimeDxeReportStatusCodeLib` sends data via a protocol rather than by linking to a particular implementation, which is why use of `PeiDxeDebugLibReportStatusCode` doesn't cause dependency cycle on build unlike `DxeRuntimeDebugLibSerialPort`.

***

The only TODO item left is the new one from @krystian-hebel' comment elsewhere:

> In light of Dasharo for Q35, possibly drop all OVMF changes

I'm not 100% sure whether to go ahead with it given that CI and OSFV use OVMF.  If it's not disruptive (using old version is fine), can be done now, otherwise the changes could be dropped on the next rebase after transitioning to Q35 target.

krystian-hebel:
> EFI vars can't work in QEMU because SMI handler fails to find CPU node which send request when the calling code is above 4GiB boundary. As a workaround, maybe could return node `0` if there is only one node.

Oh, that's what you meant, I totally misunderstood previous comment. I thought it was SMM code that somehow executed something above 4GiB. I'll see what's the best way to fix it, per-core PAE may be too expensive for SMM.

Well, I'm stupid, this won't work even for <4GiB unless it is identity mapped :facepalm:

SergiiDmytruk:
> per-core PAE may be too expensive for SMM.

64-bit version of SMM stub seems to enter long mode on every invocation:
* https://github.com/Dasharo/coreboot/blob/93a1ad17ee36b8b9286d452e13e8863aaebf8e39/src/cpu/x86/smm/smm_stub.S#L199
* https://github.com/Dasharo/coreboot/blob/93a1ad17ee36b8b9286d452e13e8863aaebf8e39/src/cpu/x86/64bit/entry64.inc#L25

krystian-hebel:
Yes, but those use constant 1:1 mappings. To analyze host code, the handler would have to map memory around RIP for each core in turn. This requires checking mode of execution the core was in when entering SMM and then parsing saved CR3 value and paging structures accordingly. (Also, in current form there would be noticeable difference in execution time depending on which core's RAX has the magic value, which technically may leak information, although limited to few predefined byte values). It is definitely possible, but would require much new code (there is almost no support for dynamically mapped memory in coreboot as of now) for little reward.

DasharoModulePkg/DasharoBootPoliciesVTd/BootPolicies.c
```diff
@@ -0,0 +1,69 @@
...
+--*/
+
+/** @file
+**/
+
+#include <Library/PcdLib.h>
+#include <Library/DebugLib.h>
+#include <Library/UefiLib.h>
+#include "Library/DasharoSystemFeaturesUiLib/DasharoSystemFeaturesHii.h"
+
```
krystian-hebel, line 18: This is missing the following:

```
#include <DasharoOptions.h>
#include <Library/UefiBootServicesTableLib.h>
#include <Library/UefiRuntimeServicesTableLib.h>
```

DasharoModulePkg/DasharoBootPoliciesVTd/BootPolicies.c
```diff
@@ -0,0 +1,69 @@
...
+EFIAPI
+InitializeBootPolicies (
+  IN EFI_HANDLE       ImageHandle,
+  IN EFI_SYSTEM_TABLE *SystemTable
+  )
+
+{
+  EFI_STATUS Status;
+  UINTN VarSize;
+  IOMMU_CONFIG *IommuConfig;
```
krystian-hebel:
```suggestion
  DASHARO_IOMMU_CONFIG *IommuConfig;
```

krystian-hebel (changes requested):

DasharoModulePkg/DasharoBootPoliciesVTd/BootPolicies.c
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/d615694de53d27e2e6ba18bc4538634633cf2a37..f4b94d7f1d4c59ad269db15034d4f9a6149afc55#diff-fedf6970b29e76db5e436bfd986fc704809f2e6caf5fd797c33bd37ed804e445R13-R17

DasharoModulePkg/DasharoBootPoliciesVTd/BootPolicies.c
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/d615694de53d27e2e6ba18bc4538634633cf2a37..f4b94d7f1d4c59ad269db15034d4f9a6149afc55#diff-fedf6970b29e76db5e436bfd986fc704809f2e6caf5fd797c33bd37ed804e445R35

krystian-hebel:
There are problems with SecureBoot, both on QEMU and NovaCustom. It seems that after enabling SB it gets back to disabled state after a reboot.

I'm also having difficulties with iPXE, there is no way to get to its shell, but this may be caused by my config and not by edk2. In any case, these are logs from both failures: [qemu-ubuntu.zip](https://github.com/user-attachments/files/15889416/qemu-ubuntu.zip)

SergiiDmytruk:
> There are problems with SecureBoot, both on QEMU and NovaCustom. It seems that after enabling SB it gets back to disabled state after a reboot.

I think it shouldn't work on `dasharo` either because https://github.com/Dasharo/edk2/pull/129/commits/055cbbe0ac73fc55fd85ca4f4da5bbe0bbe5d949#diff-f98bbf2338374177636b95ec43ff410af22e05421f313290590e74573db4cdbdR250-R253 comes from https://github.com/Dasharo/edk2/commit/2e9b6b31104fa58fee13f68918dfb8d233d2a9c8#diff-f98bbf2338374177636b95ec43ff410af22e05421f313290590e74573db4cdbdR239-R241.  @miczyg1, the intention was to check for PCD and `FirmwareUpdateMode` variable being `FALSE`, right?

SergiiDmytruk:
> I'm also having difficulties with iPXE, there is no way to get to its shell, but this may be caused by my config and not by edk2.

Looks like this is another thing broken on `dasharo` as well.  I made it so that `FirmwareUpdateMode` is initialized on boot along with all other variables, but `dasharo.ipxe` checks only existence of the variable, not its value, when it jumps to DTS.  Which is why the menu is skipped.

miczyg1:
There was this assumption that if the FUM variable exists, then we have update mode. Thus the iPXE works how it work.

macpijan:
@miczyg1 Can you also answer to:

>  @miczyg1, the intention was to check for PCD and FirmwareUpdateMode variable being FALSE, right?

I think this is correct, though. Can you explain @SergiiDmytruk  why do you suggest it should check for `FALSE`, though? The SB was meant to be disabled, when the FUM was active, right?

macpijan:
> Looks like this is another thing broken on dasharo as well. I made it so that FirmwareUpdateMode is initialized on boot along with all other variables, but dasharo.ipxe checks only existence of the variable, not its value, when it jumps to DTS. Which is why the menu is skipped.

But why does it happen? Do we have FUM set to some value initially?

Those are old changes, we have published some releases since then, I am sure we should have caught that if that would not work as expected.

SergiiDmytruk:
> Do we have FUM set to some value initially?

Yes, at the moment.
 
> Those are old changes, we have published some releases since then, I am sure we should have caught that if that would not work as expected.

https://github.com/Dasharo/DasharoModulePkg/pull/45/commits/8ecc6874ad6ca4bac4e6e5525653b86839ad794e is a new change.  Sent https://github.com/Dasharo/DasharoModulePkg/pull/48 to fix this.  It will also need bumps in EDK and coreboot to have any effect.

miczyg1:
> > There are problems with SecureBoot, both on QEMU and NovaCustom. It seems that after enabling SB it gets back to disabled state after a reboot.
> 
> I think it shouldn't work on `dasharo` either because [055cbbe#diff-f98bbf2338374177636b95ec43ff410af22e05421f313290590e74573db4cdbdR250-R253](https://github.com/Dasharo/edk2/commit/055cbbe0ac73fc55fd85ca4f4da5bbe0bbe5d949#diff-f98bbf2338374177636b95ec43ff410af22e05421f313290590e74573db4cdbdR250-R253) comes from [2e9b6b3#diff-f98bbf2338374177636b95ec43ff410af22e05421f313290590e74573db4cdbdR239-R241](https://github.com/Dasharo/edk2/commit/2e9b6b31104fa58fee13f68918dfb8d233d2a9c8#diff-f98bbf2338374177636b95ec43ff410af22e05421f313290590e74573db4cdbdR239-R241). @miczyg1, the intention was to check for PCD and `FirmwareUpdateMode` variable being `FALSE`, right?

The purpose of these checks were simple. If mere FUM var presence is enough to trigger update mode, somebody could create such variable in OS to trigger update mode even for a build where FUM was not supposed to be supported. The PCD is supposed to control the FUM availability overall.

SergiiDmytruk:
Pulled in latest `DasharoModulePkg` changes.  SB and iPXE issues seem to be gone.

krystian-hebel (approved):

mkopec:
It looks like we lost the persistent bootlogo implementation: https://github.com/Dasharo/dasharo-issues/issues/904

mkopec:
>  (possible) No pre-installed boot options? It might be OK if it's not ESP or a media.

Marking as fixed, as preinstalled boot options now appear. Also booting El Torito media was fixed by https://github.com/Dasharo/edk2/pull/129/commits/84b5cb8cbb6f8d3a48769f996c571527835fb718

mkopec (approved):
The first release from this branch has been released: https://docs.dasharo.com/variants/novacustom_v540tu/releases/

I suggest back up old `dasharo` and push this as the new `dasharo` branch

>  In light of Dasharo for Q35, possibly drop all OVMF changes

Only this point remains I think

SergiiDmytruk:
> I suggest back up old `dasharo` and push this as the new `dasharo` branch

It's probably about time to do this.

> > In light of Dasharo for Q35, possibly drop all OVMF changes
> 
> Only this point remains I think

I don't know if Q35 works in all cases where OVMF works and produces the same results (the former does lag more due to some timer implementation, I think).  I guess we might not want to rebase until moving to a newer EDK tag and adding a commit to drop OVMF changes can be done at any point, or the next rebase could get rid of those changes by not cherry-picking them.

macpijan:
What was the plan for merging this since we already have had approvals?

SergiiDmytruk:
It's a rebase.  Can't be merged, only force-pushed.

macpijan:
> It's a rebase. Can't be merged, only force-pushed.

Yes, apologies for inaccurate wording.

So we would want to proceed as described here: https://github.com/Dasharo/edk2/pull/129#pullrequestreview-2198964443

Can we do that?

Let's gather confirmation from @mkopec and @krystian-hebel and proceed if there are no problems with that.

SergiiDmytruk:
> So we would want to proceed as described here: [#129 (review)](https://github.com/Dasharo/edk2/pull/129#pullrequestreview-2198964443)

Yes, current `dasharo` should probably be renamed to `dasharo-stable202002-168` to give a hint about its upstream base (dd7523b).

> Can we do that?

Probably won't hurt to merge https://github.com/Dasharo/edk2/pull/163 first (if nobody else is going to test/review it) and then anyone who can modify protected branches can do the update.

macpijan:
@miczyg1 I will leave this one to you then.

miczyg1:
> Yes, current `dasharo` should probably be renamed to `dasharo-stable202002-168` to give a hint about its upstream base ([dd7523b](https://github.com/Dasharo/edk2/commit/dd7523b5b123de6f0730f2f2abb207f2a5c1ccd4)).

What is 168?

SergiiDmytruk:
> What is 168?

It's from the output of `git describe`: number of commits since `stable202002` tag.  `dasharo` isn't based on a tag exactly.

miczyg1:
> It's from the output of `git describe`: number of commits since `stable202002` tag. `dasharo` isn't based on a tag exactly.

Not exactly a tag, but based on stable202002 + something/ I don't think 168 introduces anything meaningful to branch name. I think `dasharo-stable202002` would be more appropriate. We will not have anything more based `stable202002` anyways

SergiiDmytruk:
> I don't think 168 introduces anything meaningful to branch name.

It's not a big deal either way.

SergiiDmytruk:
Almost forgot to provide an observation regarding `DasharoPayloadPkg` after moving capsule update changes from `DasharoPayloadPkg` to `UefiPayloadPkg` for upstreaming: some things had to be adjusted but there weren't that many conflicts and some were caused purely by formatting changes.

The takeaway is that we can reduce the difference between the two packages over time by gradually switching to libraries/DXEs of `UefiPayloadPkg` where it makes sense (small or no differences).  This can make future maintaining and upstreaming easier.  "Gradually" would mean that when we are changing some code in `DasharoPayloadPkg`, it won't hurt to compare it to corresponding  code in `UefiPayloadPkg` and evaluate if that code can be used instead.

macpijan:
@SergiiDmytruk @miczyg1 I am confused, is there any conclusion of this discussion and how to approach with finalizing the work here?

SergiiDmytruk:
I think the branches could have been updated long time ago.  Even if something else from `dasharo` is missing (as was pointed out by @miczyg1 internally in the task), so what?  Those changes will just be added to a branch with a different name.

miczyg1:
> @SergiiDmytruk @miczyg1 I am confused, is there any conclusion of this discussion and how to approach with finalizing the work here?

Nothing to finalize, except branch name changes. The rebased branch is already used in production and is a part of some releases...

miczyg1:
Changed branches:

 - current `dasharo` -> `dasharo-stable202002`
 - current `rebased` -> `dasharo`
 
 Closing this PR and removing branches: `rebased` and `rebase-base`

PR #131, opened 2024-05-18
Pick up UEFI capsules from CBMEM

SergiiDmytruk:
Corresponding coreboot changes add a CBMEM entry per capsule, quite similar to PEI's HOBs.

### DIFF
--- a/UefiPayloadPkg/BlSupportPei/BlSupportPei.c
+++ b/UefiPayloadPkg/BlSupportPei/BlSupportPei.c
@@ -752,6 +752,15 @@ BlPeiEntryPoint (
   }
 
   //
+  // Import update capsules, if there are any.
+  //
+  Status = ParseCapsules (BuildCvHob);
+  if (EFI_ERROR (Status)) {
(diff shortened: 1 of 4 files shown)


MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.inf
```diff
@@ -51,6 +52,7 @@
   DisplayUpdateProgressLib
   FileHandleLib
   UefiBootManagerLib
+  BlParseLib
```
krystian-hebel: Would it be possible to do the parsing in `BlSupportPei` and create HOB(s) with `EFI_HOB_TYPE_UEFI_CAPSULE` type there? Adding dependencies to anything in `MdeModulePkg` would increase maintenance burden.

SergiiDmytruk: Yes and that's a better place for it.  New implementation: addfc78fd93743318ab089dc613e3e10caf4bba4.

krystian-hebel (approved):

PR #134, opened 2024-05-27
Pci pmem

krystian-hebel:


macpijan:
@SergiiDmytruk Can you please take a look if you have approached this area as well when rebasing (updating) edk2?

SergiiDmytruk:
I remember those patches, but I didn't analyze the code and reasons for them much.  As I understand:
 * coreboot produces state which EDK can't handle
 * both old (no or very little pmem) and new code (some pmem) are workarounds for that
 * new code makes `if`-statement in https://github.com/Dasharo/edk2/blob/176c51dc69645857ed51f76ba9ac67b55c45c18c/MdeModulePkg/Bus/Pci/PciBusDxe/PciIo.c#L1955-L1972
    to not execute for some region of memory in case of coreboot+EDK in Qemu, because now it remains pmem

If that's correct, then this is an improvement in a sense that things are less broken in this case :)  We might want to have this coreboot<->EDK incompatibility tracked somewhere though if it's not already tracked.

krystian-hebel:
Possibly the same thing caused https://ticket.coreboot.org/issues/415 and https://ticket.coreboot.org/issues/499. In QEMU I initially had it the other way around, `CONFIG_RESOURCE_ALLOCATION_TOP_DOWN` made it ~~work~~ fail elsewhere.

SergiiDmytruk (approved):

PR #132, opened 2024-05-22
MTL graphics

mkopec:

PR #136, opened 2024-05-28
DasharoPayloadPkg/GraphicsOutputDxe: Allow FB to be at offset from BAR

mkopec:
On MTL the framebuffer is at BAR2 + some offset. Allow for this case when validating the framebuffer info received from coreboot.

Same as https://github.com/Dasharo/edk2/pull/132 but for rebased edk2.

### DIFF
--- a/DasharoPayloadPkg/GraphicsOutputDxe/GraphicsOutput.c
+++ b/DasharoPayloadPkg/GraphicsOutputDxe/GraphicsOutput.c
@@ -409,6 +409,10 @@ GraphicsOutputDriverBindingStart (
               if (Resources->AddrRangeMin == GraphicsInfo->FrameBufferBase) {
                 FrameBufferBase = Resources->AddrRangeMin;
                 break;
+              } else if (Resources->AddrRangeMin <= GraphicsInfo->FrameBufferBase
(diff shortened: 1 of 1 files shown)


mkopec:
Tested on V540TU, works correctly.

miczyg1 (approved):

DasharoPayloadPkg/GraphicsOutputDxe/GraphicsOutput.c
```diff
@@ -409,6 +409,10 @@ GraphicsOutputDriverBindingStart (
               if (Resources->AddrRangeMin == GraphicsInfo->FrameBufferBase) {
                 FrameBufferBase = Resources->AddrRangeMin;
                 break;
+              } else if (Resources->AddrRangeMin <= GraphicsInfo->FrameBufferBase
```
SergiiDmytruk: I think the new `if`-statement can just replace the old one (new condition is a superset of the old one) to make the code simpler.

mkopec: Okay, removed the old `if` in commit beba2f7540ab56537875bda1f6b8acc8bad9af5b

SergiiDmytruk (approved):

PR #138, opened 2024-06-10
.gitmodules: use gerrit mirror for cmocka

macpijan:
https://git.cryptomilk.org/projects/cmocka.git/ failed to clone too many times during the last 2-3 weeks.


macpijan:
This should be fixed as well when we fully switch to updated version of edk2: https://github.com/tianocore/edk2/blob/master/.gitmodules#L9

krystian-hebel (approved):

PR #139, opened 2024-06-12
Publish EDK's TPM eventlog that contains coreboot's entries

SergiiDmytruk:
Missing entries from coreboot's log are filled in as `0x01 0x00....` which affects expected SHA-1 PCR-2 value computed by `tpm2_eventlog`, but TCG specification says that all entries need to have identical set of digests listed in the same order, hence I've added those fake values as described by TXT spec because I don't think TCG covers such a case.

See commit messages for more details.

coreboot PR: https://github.com/Dasharo/coreboot/pull/517

### DIFF
--- a/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
+++ b/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
@@ -11,6 +11,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <IndustryStandard/Acpi.h>
 #include <IndustryStandard/PeImage.h>
 #include <IndustryStandard/TcpaAcpi.h>
+#include <IndustryStandard/Tpm2Acpi.h>
 
 #include <Guid/GlobalVariable.h>
 #include <Guid/HobList.h>
@@ -21,6 +22,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Guid/TpmInstance.h>
 #include <Guid/DeviceAuthentication.h>
 
+#include <Protocol/AcpiTable.h>
 #include <Protocol/DevicePath.h>
 #include <Protocol/MpService.h>
 #include <Protocol/VariableWrite.h>
@@ -46,6 +48,40 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/ReportStatusCodeLib.h>
 #include <Library/Tcg2PhysicalPresenceLib.h>
 #include <Library/DasharoVariablesLib.h>
+#include <Library/TpmMeasurementLib.h>
+
+#pragma pack(1)
+
+typedef struct {
+  EFI_ACPI_DESCRIPTION_HEADER Header;
+  // Flags field is replaced in version 4 and above
+  //    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
+  //    BIT16~31: Reserved
+  UINT32                      Flags;
+  UINT64                      AddressOfControlArea;
+  UINT32                      StartMethod;
+  UINT8                       PlatformSpecificParameters[12];  // size up to 12
+  UINT32                      Laml;                          // Optional
+  UINT64                      Lasa;                          // Optional
+} EFI_TPM2_ACPI_TABLE_V4;
+
+EFI_TPM2_ACPI_TABLE_V4  mTpm2AcpiTemplate = {
+  {
+    EFI_ACPI_5_0_TRUSTED_COMPUTING_PLATFORM_2_TABLE_SIGNATURE,
+    sizeof (mTpm2AcpiTemplate),
+    EFI_TPM2_ACPI_TABLE_REVISION,
+    //
+    // Compiler initializes the remaining bytes to 0
+    // These fields should be filled in in production
+    //
+  },
+  0, // BIT0~15:  PlatformClass
+     // BIT16~31: Reserved
+  0, // Control Area
+  EFI_TPM2_ACPI_TABLE_START_METHOD_TIS, // StartMethod
+};
+
+#pragma pack()
 
 #define PERF_ID_TCG2_DXE  0x3120
 
@@ -2774,6 +2810,106 @@ InstallTcg2 (
 }
 
 /**
+  Publish TPM2 ACPI table
+
+  @retval   EFI_SUCCESS     The TPM2 ACPI table is published successfully.
+  @retval   Others          The TPM2 ACPI table is not published.
+
+**/
+EFI_STATUS
+PublishTpm2 (
+  VOID
+  )
+{
+  EFI_STATUS                     Status;
+  EFI_ACPI_TABLE_PROTOCOL        *AcpiTable;
+  UINTN                          TableKey;
+  UINT64                         OemTableId;
+  EFI_TPM2_ACPI_CONTROL_AREA     *ControlArea;
+  TPM2_PTP_INTERFACE_TYPE        InterfaceType;
+
+  //
+  // Measure to PCR[0] with event EV_POST_CODE ACPI DATA.
+  // The measurement has to be done before any update.
+  // Otherwise, the PCR record would be different after event log update
+  // or the PCD configuration change.
+  //
+  TpmMeasureAndLogData(
+    0,
+    EV_POST_CODE,
+    EV_POSTCODE_INFO_ACPI_DATA,
+    ACPI_DATA_LEN,
+    &mTpm2AcpiTemplate,
+    mTpm2AcpiTemplate.Header.Length
+    );
+
+  mTpm2AcpiTemplate.Header.Revision = PcdGet8(PcdTpm2AcpiTableRev);
+  DEBUG((DEBUG_INFO, "Tpm2 ACPI table revision is %d\n", mTpm2AcpiTemplate.Header.Revision));
+
+  //
+  // PlatformClass is only valid for version 4 and above
+  //    BIT0~15:  PlatformClass
+  //    BIT16~31: Reserved
+  //
+  if (mTpm2AcpiTemplate.Header.Revision >= EFI_TPM2_ACPI_TABLE_REVISION_4) {
+    mTpm2AcpiTemplate.Flags = (mTpm2AcpiTemplate.Flags & 0xFFFF0000) | PcdGet8(PcdTpmPlatformClass);
+    DEBUG((DEBUG_INFO, "Tpm2 ACPI table PlatformClass is %d\n", (mTpm2AcpiTemplate.Flags & 0x0000FFFF)));
+  }
+
+  mTpm2AcpiTemplate.Laml = PcdGet32(PcdTpm2AcpiTableLaml);
+  mTpm2AcpiTemplate.Lasa = PcdGet64(PcdTpm2AcpiTableLasa);
+  if ((mTpm2AcpiTemplate.Header.Revision < EFI_TPM2_ACPI_TABLE_REVISION_4) ||
+      (mTpm2AcpiTemplate.Laml == 0) || (mTpm2AcpiTemplate.Lasa == 0)) {
+    //
+    // If version is smaller than 4 or Laml/Lasa is not valid, rollback to original Length.
+    //
+    mTpm2AcpiTemplate.Header.Length = sizeof(EFI_TPM2_ACPI_TABLE);
+  }
+
+  InterfaceType = PcdGet8(PcdActiveTpmInterfaceType);
+  switch (InterfaceType) {
+  case Tpm2PtpInterfaceCrb:
+    mTpm2AcpiTemplate.StartMethod = EFI_TPM2_ACPI_TABLE_START_METHOD_COMMAND_RESPONSE_BUFFER_INTERFACE;
+    mTpm2AcpiTemplate.AddressOfControlArea = PcdGet64 (PcdTpmBaseAddress) + 0x40;
+    ControlArea = (EFI_TPM2_ACPI_CONTROL_AREA *)(UINTN)mTpm2AcpiTemplate.AddressOfControlArea;
+    ControlArea->CommandSize  = 0xF80;
+    ControlArea->ResponseSize = 0xF80;
+    ControlArea->Command      = PcdGet64 (PcdTpmBaseAddress) + 0x80;
+    ControlArea->Response     = PcdGet64 (PcdTpmBaseAddress) + 0x80;
+    break;
+  case Tpm2PtpInterfaceFifo:
+  case Tpm2PtpInterfaceTis:
+    break;
+  default:
+    DEBUG((EFI_D_ERROR, "TPM2 InterfaceType get error! %d\n", InterfaceType));
+    break;
+  }
+
+  CopyMem (mTpm2AcpiTemplate.Header.OemId, PcdGetPtr (PcdAcpiDefaultOemId), sizeof (mTpm2AcpiTemplate.Header.OemId));
+  OemTableId = PcdGet64 (PcdAcpiDefaultOemTableId);
+  CopyMem (&mTpm2AcpiTemplate.Header.OemTableId, &OemTableId, sizeof (UINT64));
+  mTpm2AcpiTemplate.Header.OemRevision      = PcdGet32 (PcdAcpiDefaultOemRevision);
+  mTpm2AcpiTemplate.Header.CreatorId        = PcdGet32 (PcdAcpiDefaultCreatorId);
+  mTpm2AcpiTemplate.Header.CreatorRevision  = PcdGet32 (PcdAcpiDefaultCreatorRevision);
+
+  //
+  // Construct ACPI table
+  //
+  Status = gBS->LocateProtocol (&gEfiAcpiTableProtocolGuid, NULL, (VOID **) &AcpiTable);
+  ASSERT_EFI_ERROR (Status);
+
+  Status = AcpiTable->InstallAcpiTable (
+                        AcpiTable,
+                        &mTpm2AcpiTemplate,
+                        mTpm2AcpiTemplate.Header.Length,
+                        &TableKey
+                        );
+  ASSERT_EFI_ERROR (Status);
+
+  return Status;
+}
+
+/**
   The driver's entry point. It publishes EFI Tcg2 Protocol.
 
   @param[in] ImageHandle  The firmware allocated handle for the EFI image.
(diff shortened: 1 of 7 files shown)


UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
```diff
@@ -730,6 +766,49 @@ ParseTPMPPIInfo (
...
+  @param  Base          Pointer the start of the log
+  @param  Size          Size of the log
+
+  @retval RETURN_SUCCESS     Successfully found TPM log.
+  @retval RETURN_NOT_FOUND   Failed to find TPM log.
+
+**/
+RETURN_STATUS
+EFIAPI
+ParseTPMLog (
```
miczyg1: This actually does not do any parsing.  Maybe `LocateTpmLog`?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -2797,5 +2978,11 @@ DriverEntry (
   Status = InstallTcg2 ();
   DEBUG ((EFI_D_INFO, "InstallTcg2 - %r\n", Status));
 
+  //
+  // Set TPM2 ACPI table
+  //
+  Status = PublishTpm2 ();
```
miczyg1, line 3100: Okay we publish TPM2 table, but we haven't made equivalent changes to publish TCPA table for TPM1.2 case.

When TPM1.2 is detected, it doesn't use Tcg2Dxe modules, so I wonder how everything will apply to TPM1.2 case.

miczyg1 (changes requested):

UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
SergiiDmytruk: All functions in [UefiPayloadPkg/Include/Library/BlParseLib.h](https://github.com/Dasharo/edk2/pull/139/files#diff-513d54e33dc93f5ffff51bfca23f56624568863b93dd10f966014b99455bfb46) follow `Parse*` naming scheme despite not doing parsing, I kept new code consistent with the rest.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
SergiiDmytruk: EDK is inconsistent :)  Unlike Tcg2Dxe which left installation of ACPI table to Tcg2Smm, TcgDxe publishes ACPI table on its own.  Weirdly enough TcgSmm also publishes some ACPI table which it reads from FV, not sure why.

SergiiDmytruk (commented):
Force-pushed to get rid of a stray formatting fix in `CbParseLib.c`.

UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
miczyg1: Okay.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
miczyg1: @SergiiDmytruk any update here?

SergiiDmytruk: What for?  There are no changes for TPM1.2 because they aren't needed: TcgDxe already publishes the log.

Just checked it again to be sure and it does, although when coreboot supports both TPM versions TPM2.0 log format is used and that makes EDK to fail importing the log.  If we ship boards that support both TPM versions either need to pick log version dynamically in  coreboot or import truncated SHA-256 hashes from TPM2 log in EDK (because there are no SHA-1 hashes).  Otherwise event log will miss coreboot part when TPM1 is in use.

miczyg1: >  TcgDxe already publishes the log

And it doesn't need any modifications? 

> If we ship boards that support both TPM versions either need to pick log version dynamically in coreboot or import truncated SHA-256 hashes from TPM2 log in EDK (because there are no SHA-1 hashes). 

I would expect coreboot to create an appropriate log depending on detected TPM

SergiiDmytruk: > And it doesn't need any modifications?

It's just SHA-1 all the time for TPM1, so log entries from coreboot and EDK already align with each other unless you want to drop trailing zeros from events coming from coreboot (not done for TPM2 either).

> I would expect coreboot to create an appropriate log depending on detected TPM

You brought that up during review [here](https://review.coreboot.org/c/coreboot/+/68747/comment/cbfd1fd6_db1c9692/).  I guess there could be a separate `TPM_LOG_TCG` that picks log format automatically.  On Talos TPM and log versions didn't align with each other, I think that's why log format needed to be independent.

miczyg1: OK, I get it, that sometimes you may want to force a log format.

So what will happen when coreboot creates the TPM2-compliant log format for TPM1.2? Will EDK2 actually use the same format in TcgDxe? This is my main concern

SergiiDmytruk: > So what will happen when coreboot creates the TPM2-compliant log format for TPM1.2?

coreboot's event log won't be imported.  Also because TcgDxe doesn't create its own Event ID entry, there will be none.

miczyg1: Sorry, but I still don't understand your previous statement:
`What for? There are no changes for TPM1.2 because they aren't needed: TcgDxe already publishes the log.`
Together with this: `coreboot's event log won't be imported. Also because TcgDxe doesn't create its own Event ID entry, there will be none.` I'm unable to be convinced that there are no modifications required

SergiiDmytruk: No modifications required for making EDK's TPM1.2 event log available to an OS.  Changing what's published requires modifications here or more likely in coreboot if we want to support those scenarios (picking TPM1 or TPM2 dynamically).

miczyg1: >  if we want to support those scenarios (picking TPM1 or TPM2 dynamically).

That was the goal of the whole effort of unifying TPM drivers and eventlogs. To get the right driver talk to TPM and have coreboot create an appropriate log for the detected TPM, rather than rely on fixed CONFIG.

While this PR addresses the problem of the missing log entries from coreboot, I would also want to see it working correctly in the dynamic scenarios.
@macpijan cc

I can close this thread if I see some next steps taken in that direction (i.e., tasks/PRs).

SergiiDmytruk: Looks like I almost have the changes for enabling dynamic log selection in coreboot.  It wasn't hard to do them, but original code for registering probing functions (not the one which landed in upstream) didn't work in some stages and the log was missing stuff.  Now things work for TPM1 fine but there is somehow no log in OS for TPM2.  Can send a coreboot PR after sorting it out.

SergiiDmytruk: > Now things work for TPM1 fine but there is somehow no log in OS for TPM2.

Seems like an issue on Q35 or with Qemu/swtpm because the same happens without new changes and everything works fine on APU.

I've added the changes to https://github.com/Dasharo/coreboot/pull/517.  There is a side-effect when new option is on: probing for TPM in basically every stage which needs to know algorithm for hashing data because the algorithm depends on TPM now.

macpijan:
@miczyg1 any more comments here?

miczyg1:
> @miczyg1 any more comments here?

Actually not anymore. I would like to give this PR and [this](https://github.com/Dasharo/coreboot/pull/517) a try on some HW that has a socketable TPM 1.2 and TPM 2.0. https://github.com/Dasharo/coreboot/pull/517

SergiiDmytruk:
Added a fix to make measurements of Dasharo variables reproducible ([c0078b2](https://github.com/Dasharo/edk2/pull/139/commits/c0078b2fe2c91e5aeaa858de6bee42914540632c)) here because it's kind of related.  https://github.com/Dasharo/DasharoModulePkg has the issue as well, but don't know if we need to fix it there.

miczyg1:
> https://github.com/Dasharo/DasharoModulePkg has the issue as well, but don't know if we need to fix it there.

We should, there are still releases coming that will use DasharoModulePkg submodule in EDKII

SergiiDmytruk:
Rebased.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
```diff
@@ -307,7 +307,7 @@ SyncPcrAllocationsAndPcrMask (
   {
     DEBUG ((DEBUG_INFO, "TpmActivePcrBanks & Tpm2PcrMask = 0x%08x\n", (TpmActivePcrBanks & Tpm2PcrMask)));
     DEBUG ((DEBUG_INFO, "TpmActivePcrBanks & BiosHashAlgorithmBitmap = 0x%08x\n", (TpmActivePcrBanks & BiosHashAlgorithmBitmap)));
-    NewTpmActivePcrBanks  = TpmActivePcrBanks;
+    NewTpmActivePcrBanks  = TpmHashAlgorithmBitmap;
```
miczyg1: This will break the feature of changing PCR banks via menu. Am I right? It always forces to use the single bank defined by PCD at coreboot build time.

miczyg1 (changes requested):
Tested it on various platforms and it works well.

However I am worried about the flexibility of the solution. We are forcing only one hashing algorithm on both sides: coreboot and EDK2. It will break the PCR bank switching feature. I would expect both coreboot and EDK2 to provide measurement using all hashes that are currently enabled in TPM.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
SergiiDmytruk: That is correct, but coreboot never extends more than a single bank and EDK2 was constrained to do the same to align the two parts of the firmware as a simpler option: https://github.com/Dasharo/dasharo-issues/issues/982

miczyg1: Yes, and that should also change per my comment: https://github.com/Dasharo/edk2/pull/139#pullrequestreview-2464945893

miczyg1:
Rebased on top of recent dasharo

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
SergiiDmytruk: Making coreboot use active PCRs is not that easy, so I just removed constraining set of digests in EDK from here and https://github.com/Dasharo/coreboot/pull/517.

miczyg1: So now EDK2 and coreboot can extend different digests...

Worst case scenario is that the TPM does not have SHA256 bank enabled (which is used by default in coreboot), so some part of the log will be lost.

Not so bad case is that SHA256 bank is enabled and SHA1 bank is enabled, but coreboot will extend only SHA256. EDK2 will extend both SHA1 and SHA256. The final eventlog should reproduce both PCR sets properly, although there will be less SHA1 measurements. Am I correct?

SergiiDmytruk: > So now EDK2 and coreboot can extend different digests...

Which is how it works on `dasharo` at the moment.

> Worst case scenario is that the TPM does not have SHA256 bank enabled (which is used by default in coreboot), so some part of the log will be lost.

Yes, event log won't have any real digests from coreboot.

> Not so bad case is that SHA256 bank is enabled and SHA1 bank is enabled, but coreboot will extend only SHA256. EDK2 will extend both SHA1 and SHA256. The final eventlog should reproduce both PCR sets properly, although there will be less SHA1 measurements. Am I correct?

PCRs will match values computed from event log only for a single bank which is extended by both coreboot and EDK2.  Event log entries must have identical set of digests and this PR adds missing digests filled as described by TXT spec (`01000000...` in hex).

Basically, this leaves it to the user which should enable SHA256 in EDK2 for measured boot to work and disable all other banks to not see mismatches.  Although I think that might be the default state for many TPMs.  Still strictly an improvement compared to the current state when there is no chance of measured boot to work.

miczyg1: > Basically, this leaves it to the user which should enable SHA256 in EDK2 for measured boot to work and disable all other banks to not see mismatches.

Exactly. Not perfect, but we could get it fixed later...

>  Although I think that might be the default state for many TPMs

SHA256 must be enabled by default on TPM2.

> Still strictly an improvement compared to the current state when there is no chance of measured boot to work.

It was an improvement from the beginning, no denying it.

What exactly is the difficulty on coreboot side to have agile logging?

SergiiDmytruk: > Exactly. Not perfect, but we could get it fixed later...

Yep, there is already https://github.com/Dasharo/dasharo-issues/issues/982 for it.

> What exactly is the difficulty on coreboot side to have agile logging?

It's not trivial code-wise to switch the logging and measurement code from one hash to several determined at run-time.  My latest attempt can be seen at https://github.com/Dasharo/coreboot/compare/59165487cd34b9f3a2d9f41b4fa28695fdc1b47b..all-tpm-banks.  I thought https://github.com/Dasharo/coreboot/commit/c5bb9b4215b940e0cbb784d52e779f6eefe2575e just needed a rebase, but it had set of hashes fixed at build-time.  I ended up also taking changes from some of my stashes and adding querying list of active PCRs.  Things left to do:
 - update TPM1.2 log to use new API (my rebase added an extra file instead)
 - fix either extending or copying pre-RAM eventlog, something goes wrong there
 - fix querying set of active PCRs which had some trouble before ramstage and possibly broke bootblock (APU refused to boot and gave no serial output)

It looked like it worked fine in QEMU with arbitrary set of banks, but I think it had trouble extending PCRs from event log cache at the time which I noticed only with APU.  So nothing from before ramstage was actually measured and that's what needs debugging and fixing.

miczyg1: So it seems to boil down to pre-RAM measurements based on these:

`copying pre-RAM eventlog, something goes wrong there`

`It looked like it worked fine in QEMU with arbitrary set of banks, but I think it had trouble extending PCRs from event log cache at the time which I noticed only with APU. So nothing from before ramstage was actually measured and that's what needs debugging and fixing.`

The available space in CAR (cache-as-RAM) is limited. Apu2 may have the most scarce space in cache for that TBH. That is where the problem may live.

The priority for migration should be as follows:
1. coreboot uses the space defined in car.ld for pre-RAM logs until cbmem is ready:

```
#if CONFIG(TPM_MEASURED_BOOT)
	/* Vboot measured boot TPM log measurements.
	 * Needs to be transferred until CBMEM is available */
	TPM_LOG(., 2K)
#endif
```

It is only 2K, with multiple digests, we may quickly run out of space.

2. As soon as cbmem is up, migrate the pre-RAM logs from `TPM_LOG` in cache to cbmem entry.
3. For stages which have cbmem and cbmem is already up, use the cbmem area to store digests (as if they would be logged in ramstage). It should be possible by the end of romstage I believe. So logging postcar stage measurement and later should already be possible with eventlog in cbmem. Up to that point how many measurements can we have? bootblock CRTM, romstage, FSP or AGESA (this one may be measured multiple times as it is access multiple times), for vboot it may be additionally GBB, HWID, boot mode etc. I wrost case 10 measurements?

If you are still up to some debugging, might be worth considering to do the testing on more modern platform with bigger cache size. Then go back to apu2, as it may be possible to run out of cache for code execution (stack and other data).

SergiiDmytruk: Right, there were messages about the log being full.  I thought it could be a bug, didn't get to checking that it's actual lack of space.

Thanks for the suggestions, I'll do more debugging later and it should be working at some point.  Those changes are coreboot-only though, so this PR (and likely related coreboot PR) might as well be merged.  Using all PCR banks in coreboot will take separate review effort due to significant amount of changes.

miczyg1: > Right, there were messages about the log being full. I thought it could be a bug, didn't get to checking that it's actual lack of space.

I have seen this on many platforms, so I doubt a bit that it is really the lack of space...

> Those changes are coreboot-only though, so this PR (and likely related coreboot PR) might as well be merged. 

Yes, I will take a look once again ASAP

mkopec:
Tested on novacustom, we now get valid PCR0 reconstruction:

```
HSI-1
✔ BIOS firmware updates:         Enabled
✔ csme18 override:               Locked
✔ csme18 v0:18.0.5.2040:         Valid
✔ Platform debugging:            Disabled
✔ SPI lock:                      Enabled
✔ Supported CPU:                 Valid
✔ TPM empty PCRs:                Valid
✔ TPM v2.0:                      Found
✔ UEFI bootservice variables:    Locked
✔ UEFI platform key:             Valid
✘ MEI manufacturing mode:        Unlocked
✘ SPI write:                     Enabled
✘ SPI BIOS region:               Unlocked

HSI-2
✔ IOMMU:                         Enabled
✔ Platform debugging:            Locked
✔ TPM PCR0 reconstruction:       Valid
✘ Intel BootGuard ACM protected: Invalid
✘ Intel BootGuard:               Disabled
✘ Intel BootGuard OTP fuse:      Unlocked

HSI-3
✔ CET Platform:                  Supported
✔ Pre-boot DMA protection:       Enabled
✔ Suspend-to-idle:               Enabled
✔ Suspend-to-ram:                Disabled

HSI-4
✔ SMAP:                          Enabled
✘ Encrypted RAM:                 Not supported

Runtime Suffix -!
✔ fwupd plugins:                 Untainted
✔ Linux swap:                    Encrypted
✔ Linux kernel:                  Untainted
✘ CET OS Support:                Not supported
✘ Linux kernel lockdown:         Disabled
✘ UEFI secure boot:              Disabled
```

mkopec (approved):

miczyg1 (approved):

PR #141, opened 2024-06-14
Serial issues fixes

miczyg1:
FIex the problem where the serial console redirection would be truly disabled after 2nd reset.

The BS->ConnectController was mainly added for external PS/2 Keyboard detection and input reliability. As external PS/2 is is not used in our lab on any board (laptops have the PS/2 always on), drop the ConnectController call.

### DIFF
--- a/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformConsole.c
+++ b/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformConsole.c
@@ -329,9 +329,6 @@ PrepareLpcBridgeDevicePath (
     return Status;
   }
 
-  /* Start the drivers for all child devices of this controller/device */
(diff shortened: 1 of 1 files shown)


miczyg1:
I also went ahead and changed the prompts per customer request:

image

macpijan:
> The BS->ConnectController was mainly added for external PS/2 Keyboard detection and input reliability.

Can it break something with PS/2 on e.g. MSI?

miczyg1:
> > The BS->ConnectController was mainly added for external PS/2 Keyboard detection and input reliability.
> 
> Can it break something with PS/2 on e.g. MSI?

PS/2 doesn't work reliably yet on MSI and still needs some work anyway

macpijan (approved):

PR #145, opened 2024-06-24
DasharoPayloadPkg: publish ESRT with an entry for coreboot

SergiiDmytruk:
CbParseLib implementation of BlParseLib was extended to allow querying CB_TAG_FW_INFO tag from coreboot table.

And BlSupportDxe was made to use this information to publish a single-element ESRT.  Later on ESRT will likely be constructed by EsrtFmpDxe from data provided by FMP instances, but until then this should be good enough for viewing this information in an OS.

coreboot part: https://github.com/Dasharo/coreboot/pull/527

### DIFF
--- a/DasharoPayloadPkg/BlSupportDxe/BlSupportDxe.c
+++ b/DasharoPayloadPkg/BlSupportDxe/BlSupportDxe.c
@@ -226,6 +226,12 @@ BlDxeEntryPoint (
   EFI_HOB_GUID_TYPE                       *GuidHob;
   SYSTEM_TABLE_INFO                       *SystemTableInfo;
   EFI_PEI_GRAPHICS_INFO_HOB               *GfxInfo;
+  EFI_GUID                                FwGuid;
+  UINT32                                  FwVersion;
+  UINT32                                  FwLsv;
+  UINT32                                  FwSize;
+  EFI_SYSTEM_RESOURCE_TABLE               *Esrt;
+  EFI_SYSTEM_RESOURCE_ENTRY               *Esre;
 
   Status = EFI_SUCCESS;
   //
@@ -270,6 +276,25 @@ BlDxeEntryPoint (
     Status = PcdSet32S (PcdSetupVideoHorizontalResolution, GfxInfo->GraphicsMode.HorizontalResolution);
     ASSERT_EFI_ERROR (Status);
     Status = PcdSet32S (PcdSetupVideoVerticalResolution, GfxInfo->GraphicsMode.VerticalResolution);
+    ASSERT_EFI_ERROR (Status);
+  }
+
+  Status = ParseFwInfo (&FwGuid, &FwVersion, &FwLsv, &FwSize);
+  if (!EFI_ERROR (Status)) {
+    Esrt = AllocateZeroPool (sizeof (EFI_SYSTEM_RESOURCE_TABLE) + sizeof (EFI_SYSTEM_RESOURCE_ENTRY));
+    ASSERT (Esrt != NULL);
+
+    Esrt->FwResourceVersion  = EFI_SYSTEM_RESOURCE_TABLE_FIRMWARE_RESOURCE_VERSION;
+    Esrt->FwResourceCount    = 1;
+    Esrt->FwResourceCountMax = 1;
+
+    Esre = (EFI_SYSTEM_RESOURCE_ENTRY *)&Esrt[1];
+    CopyMem (&Esre->FwClass, &FwGuid, sizeof (EFI_GUID));
+    Esre->FwType                   = ESRT_FW_TYPE_SYSTEMFIRMWARE;
+    Esre->FwVersion                = FwVersion;
+    Esre->LowestSupportedFwVersion = FwLsv;
+
+    Status = gBS->InstallConfigurationTable (&gEfiSystemResourceTableGuid, Esrt);
     ASSERT_EFI_ERROR (Status);
   }
 
--- a/DasharoPayloadPkg/Library/CbParseLib/CbParseLib.c
+++ b/DasharoPayloadPkg/Library/CbParseLib/CbParseLib.c
@@ -1024,3 +1024,42 @@ ParseBootLogo (
 
   return RETURN_SUCCESS;
 }
+
+/**
+  Parse firmware information passed in by coreboot
+
+  @param  Guid     Kind of the firmware.
+  @param  Version  Current version.
+  @param  Lsv      Lowest supported version.
+  @param  Size     Firmware size in bytes.
+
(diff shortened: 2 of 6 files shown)


miczyg1:
> And BlSupportDxe was made to use this information to publish a single-element ESRT. Later on ESRT will likely be constructed by EsrtFmpDxe from data provided by FMP instances, but until then this should be good enough for viewing this information in an OS.

Can we really consider https://github.com/Dasharo/dasharo-issues/issues/802 complete with this PR? ESRT will be generated differently later, so it feels kind of weird. Or what would be the plan?

SergiiDmytruk:
> Can we really consider [Dasharo/dasharo-issues#802](https://github.com/Dasharo/dasharo-issues/issues/802) complete with this PR? ESRT will be generated differently later, so it feels kind of weird. Or what would be the plan?

Probably, because the task is about parsing that information and the same coreboot's data will be used later but in a different place.  By the way, BlSupportDxe changes won't even need to be reverted as `InstallConfigurationTable()` updates an already installed table.  Alternatively, could wait until FMP is enabled and rebase this PR then.

DasharoPayloadPkg/BlSupportDxe/BlSupportDxe.c
```diff
@@ -273,6 +278,25 @@ BlDxeEntryPoint (
...
+    Esrt = AllocateZeroPool (sizeof (EFI_SYSTEM_RESOURCE_TABLE) + sizeof (EFI_SYSTEM_RESOURCE_ENTRY));
+    ASSERT (Esrt != NULL);
+
+    Esrt->FwResourceVersion  = EFI_SYSTEM_RESOURCE_TABLE_FIRMWARE_RESOURCE_VERSION;
+    Esrt->FwResourceCount    = 1;
+    Esrt->FwResourceCountMax = 1;
+
+    Esre = (EFI_SYSTEM_RESOURCE_ENTRY *)&Esrt[1];
+    CopyMem (&Esre->FwClass, &FwGuid, sizeof (EFI_GUID));
+    Esre->FwType                   = ESRT_FW_TYPE_SYSTEMFIRMWARE;
```
krystian-hebel, line 293: Is this something we will want to pass from coreboot as well? Or leave it for FMP to worry about?

DasharoPayloadPkg/Library/CbParseLib/CbParseLib.c
```diff
@@ -991,3 +991,39 @@ ParseCapsules (
...
+  OUT UINT32    *Lsv
+  )
+{
+  struct lb_efi_fw_info  *FwInfo;
+
+  if (Guid == NULL || Version == NULL || Lsv == NULL) {
+    return RETURN_INVALID_PARAMETER;
+  }
+
+  FwInfo = FindCbTag (CB_TAG_FW_INFO);
```
krystian-hebel, line 1055: At some point we will need something like `FindNextCbTag`, I believe it would be useful in some parts of existing code, but this isn't something to worry about just yet.

krystian-hebel (commented):
I agree with Miczyg that this isn't too useful without any FMP. It provides a nice way of testing, but I think it shouldn't be merged in this form.

DasharoPayloadPkg/BlSupportDxe/BlSupportDxe.c
SergiiDmytruk: Possible values are:
* `ESRT_FW_TYPE_UNKNOWN`
* `ESRT_FW_TYPE_SYSTEMFIRMWARE`
* `ESRT_FW_TYPE_DEVICEFIRMWARE`
* `ESRT_FW_TYPE_UEFIDRIVER`

It seems like coreboot won't need anything other than `ESRT_FW_TYPE_SYSTEMFIRMWARE`.

krystian-hebel: I'm not so sure. At some point we may want to track ME version, and if it is disabled by coreboot, edk2 may no longer be able to communicate with it to ask for its version (if this is how it's done, I don't know ME to well). I think it should be of type `DEVICEFIRMWARE`, because even though it is stored on the same flash device as main firmware, it has to be flashed in a different way.

EC also is `DEVICEFIRMWARE`. @mkopec since you've worked with it a lot, do you know if there is a possible scenario in which coreboot is able to obtain EC firmware version, but payload/OS isn't?

SergiiDmytruk: Separation for capsules type seems to be based more on whether BIOS is locked or not (`ProcessCapsules()` is called first when it's unlocked which is when system firmware and devices without embedded drivers are updated and second time after applying the lock; although this separation is less applicable when coreboot is used) rather than on anything else.

Adding more fields isn't a problem, but type alone probably won't help much because there will be different kinds of ECs.  We'll likely have multiple FMP implementations each supporting a set of GUIDs that it recognizes and knows how to flash with ability to derive necessary information from a GUID.  We should also be able to add new fields describing firmware in the future and detect struct version by the size, or could add a version field from the start.

mkopec: > EC also is DEVICEFIRMWARE. @mkopec since you've worked with it a lot, do you know if there is a possible scenario in which coreboot is able to obtain EC firmware version, but payload/OS isn't?

I don't believe so, right now we don't hide or lock down any EC interfaces so everything should be available to the payload.

DasharoPayloadPkg/BlSupportDxe/BlSupportDxe.c
```diff
@@ -226,6 +226,11 @@ BlDxeEntryPoint (
   EFI_HOB_GUID_TYPE                       *GuidHob;
   SYSTEM_TABLE_INFO                       *SystemTableInfo;
   EFI_PEI_GRAPHICS_INFO_HOB               *GfxInfo;
+  EFI_GUID                                FwGuid;
```
krystian-hebel, line 229: @SergiiDmytruk please extract changes to `BlSupportDxe` to separate commit, we will drop it eventually so having dedicated commit would make it easier.

SergiiDmytruk: a5428a73362b2d25a1c220a8582f3a80eae04013 709130a9b456ed4e3836a269a8fa7239fd73b1ec

Also added firmware size to information passed from coreboot.

krystian-hebel (approved):
Approved, but let's keep it on separate branch for now.

krystian-hebel:
Further work on #147 (better branch name), commits from this PR are already there.

PR #148, opened 2024-07-01
UefiPayloadPkg/UefiPayloadPkg.dsc: Move out the PCD settings from UiApp scope

miczyg1:
Setting the PCDs only in the UiApp scope does not seem not to be propagated to the DasharoModulesPkg libraries when FixedPcdGetBool is used with those PCDs. Move away the PCD settings to the global PCD section in the DSC.


miczyg1:
@macpijan with this the missing menus/options such as Serial Port are visible.

PR #151, opened 2024-07-11
AuthVariableLib: Set SB to enabled during transition to USER_MODE

miczyg1:
When the system is booted, the AuthVariableLibInitialize runs before SecureBootDefaultKeysDxe. When the system is booted for the first time, AuthVariableLibInitialize will not set SB to enabled state, because there are no keys yet. So the SecureBootEnable does not need to be set depending on PCD, even shouldn't. It is the SecureBootDefaultKeysDxe task to set the proper SB state when system boots with default settings. The PCD is still used there and in the UI when restoring default settings.

Quick tests:

 When SB Default State PCD is `FALSE`:
 1. Secure Boot state on first boot (SMMSTORE region empty) - Disabled - PASS
 2. Secure Boot state when defaults are restored with F9 in UI - Disabled - PASS
     - enter SB menu and enable SB, reboot
     - enter SB menu and press F9, save and reboot
     - enter SB menu and confirm the SB state to be equal to the PCD value
 3. Secure Boot state when keys are enrolled from OS - Enabled - PASS
    - enter SB menu and erase all keys, reboot
    - boot Ubuntu and provision keys, e.g. http://jk.ozlabs.org/docs/sbkeysync-maintaing-uefi-key-databases/
    - reboot to setup UI and confirm SB state
 4. Booting Ubuntu with SB:
    - Booting standard shim+grub - Access Denied - PASS
    - Booting custom signed grub - PASS (but did not boot OS due to shim_lock protocol not found, kind of expected when shim was omitted)
 
 When SB Default State PCD is `TRUE`:
 1. Secure Boot state on first boot (SMMSTORE region empty) - Enabled - PASS
 2. Secure Boot state when defaults are restored with F9 in UI - Enabled - PASS
     - enter SB menu and disable SB, reboot
     - enter SB menu and press F9, save and reboot
     - enter SB menu and confirm the SB state to be equal to the PCD value
 3. Secure Boot state when keys are enrolled from OS - Enabled - PASS
    - enter SB menu and erase all keys, reboot
    - boot Ubuntu and provision keys, e.g. http://jk.ozlabs.org/docs/sbkeysync-maintaing-uefi-key-databases/
    - reboot to setup UI and confirm SB state
 4. Booting Ubuntu with SB:
    - Booting standard shim+grub - Access Denied - PASS
    - Booting custom signed grub - PASS (but did not boot OS due to shim_lock protocol not found, kind of expected when shim was omitted)
  
Questions:

1. ~~Resetting Secure Boot keys to defaults causes the SB state to become enabled, but it can also set the state to disabled if we want to. Is it the behavior we want?~~ Done, now when resetting the keys, Enable Secure Boot checkbox remains unchanged. Only the Current Secure Boot State changes from Disabled to Enabled, but there is now way around it. SecureBoot variable becomes locked after enrolling PK.
2. Pressing F9 in SB menu restores the Secure Boot enable state only to the PCD value, it does not restore the keys to defaults, but it could if we want to. Is it the behavior we want?

### DIFF
--- a/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
+++ b/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
@@ -4165,13 +4165,28 @@ ON_EXIT:
 STATIC EFI_STATUS
 EFIAPI
 KeyEnrollReset (
-  VOID
+  IN SECUREBOOT_CONFIGURATION        *ConfigData
   )
 {
   EFI_STATUS  Status;
   UINT8       SetupMode;
+  BOOLEAN     SecureBootEnable;
+  UINTN       DataSize;
 
-  Status = EFI_SUCCESS;
+  DataSize = sizeof (SecureBootEnable);
+  Status = gRT->GetVariable(
+                EFI_SECURE_BOOT_ENABLE_NAME,
+                &gEfiSecureBootEnableDisableGuid,
+                NULL,
+                &DataSize,
+                &SecureBootEnable
+                );
+
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Cannot read SecureBootEnable variable: %r\n", Status));
+    /* Get the state from the from data if we failed to read the variable */
+    SecureBootEnable = ConfigData->AttemptSecureBoot;
+  }
 
   Status = SetSecureBootMode (CUSTOM_SECURE_BOOT_MODE);
   if (EFI_ERROR(Status)) {
@@ -4256,6 +4271,20 @@ KeyEnrollReset (
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "Cannot enroll PK: %r\n", Status));
     goto clearKEK;
+  }
+
+  /* 
+   * If Secure Boot was disabled before resetting the keys, don't change its state.
+   * Enrolling PK would enable Secure Boot automatically.
+   */
+  if (SecureBootEnable == SECURE_BOOT_DISABLE) {
+    Status = SaveSecureBootVariable (SECURE_BOOT_DISABLE);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "Cannot set Secure Boot state to SECURE_BOOT_DISABLE\n"
+        ));
+    }
   }
 
   Status = SetSecureBootMode (STANDARD_SECURE_BOOT_MODE);
@@ -4969,7 +4998,7 @@ SecureBootCallback (
                            &UserSelection
                            );
       if (UserSelection == EfiHiiPopupSelectionYes) {
-        Status = KeyEnrollReset ();
+        Status = KeyEnrollReset (IfrNvData);
       }
       //
       // Update secure boot strings after key reset
--- a/SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c
+++ b/SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c
@@ -205,9 +205,13 @@ AuthVariableLibInitialize (
     }
   } else if (mPlatformMode == USER_MODE) {
     //
-    // "SecureBootEnable" not exist, initialize it in USER_MODE.
+    // "SecureBootEnable" not exist, initialize it in USER_MODE. Can't use
+    // PcdSecureBootDefaultEnable here, because it will prevent enabling
+    // Secure Boot from OS when transitioning from SETUP_MODE to USER_MODE.
+    // PcdSecureBootDefaultEnable is used when settings are reset in UI and in
+    // the SecureBootDefaultKeysDxe where the default keys are being restored.
     //
-    SecureBootEnable = FixedPcdGet8 (PcdSecureBootDefaultEnable);
+    SecureBootEnable = SECURE_BOOT_ENABLE;
     Status = AuthServiceInternalUpdateVariable (
                EFI_SECURE_BOOT_ENABLE_NAME,
                &gEfiSecureBootEnableDisableGuid,


miczyg1:
@macpijan ?

macpijan:
> Resetting Secure Boot keys to defaults causes the SB state to become enabled, but it can also set the state to disabled if we want to. Is it the behavior we want?

Is option "do not change whatever has been set prior to resetting" on the table?

Otherwise, set the default for platform, if not possible, keep disabled?

> Pressing F9 in SB menu restores the Secure Boot enable state only to the PCD value, it does not restore the keys to defaults, but it could if we want to. Is it the behavior we want?

I'd vote to keep it as is. If F9 would reset keys as well, the option of restoring default keys would be questionable to exist in the first place, if functionally equivalent to F9.

miczyg1:
> > Resetting Secure Boot keys to defaults causes the SB state to become enabled, but it can also set the state to disabled if we want to. Is it the behavior we want?
> 
> Is option "do not change whatever has been set prior to resetting" on the table?

Exactly this. Do we want the "reset keys to default" change the Secure Boot state or not?

macpijan:
> Exactly this. Do we want the "reset keys to default" change the Secure Boot state or not?

Then, I'd say it should not change, and keep whatever was set before. If user wants to change, they should do it.

macpijan:
@miczyg1 We can proceed with the decisions discussed here.

krystian-hebel:
> Exactly this. Do we want the "reset keys to default" change the Secure Boot state or not?

I'm assuming we aren't talking about `SecureBoot`, but `SecureBootEnable` variable. The former depends on keys that were enrolled during boot, the latter is just a flag that enables `SecureBoot` if and only if other requirements are met, which are well defined in the spec. `SecureBoot` isn't NV, it only lives in RAM, so there is nothing to change there. However, `SecureBoot` should change its value on "Platform Specific PKpub Clear" to disabled, which makes sense in case it can be changed without the reset - otherwise it would be possible to boot with SB reported as enabled even though keys were changed along the way.

Summing up, `SecureBootEnable` may retain its value, but `SecureBoot` must be cleared (in a sticky way until the next reboot) when keys are reset. This should be reflected in efivars after keys are enrolled from within OS: boot with SB enabled and custom keys enrolled (user mode), change the keys by writing new ones signed with old keys, check value of `SecureBoot`.

Please add some description to the commit message, like the firs paragraph of this PR description. Also, how does it relate to [this](https://github.com/Dasharo/edk2/commit/4c6a12365f15cf54b649e1a8cd311ddbf66e4918)?

miczyg1:
> However, SecureBoot should change its value on "Platform Specific PKpub Clear" to disabled, which makes sense in case it can be changed without the reset - otherwise it would be possible to boot with SB reported as enabled even though keys were changed along the way.

Yes, when keys are erased, SB is being disabled. That's perfectly understandable.

> Summing up, SecureBootEnable may retain its value, but SecureBoot must be cleared (in a sticky way until the next reboot) when keys are reset. 

Yes, I think we want to retain SecureBootEnable value, because if someone had all keys enrolled and SB disbalde, resetting the keys would enable SB automatically. We enforce reboot/reset after every change in the menu, so SecureBoot variable will always have a proper value after reboot.

miczyg1:
Now when resetting the keys, Enable Secure Boot checkbox remains unchanged. Only the Current Secure Boot State changes from Disabled to Enabled, but there is now way around it. SecureBoot variable becomes locked after enrolling PK.

SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
```diff
@@ -4165,13 +4165,28 @@ LoadSignatureData (
...
+                &gEfiSecureBootEnableDisableGuid,
+                NULL,
+                &DataSize,
+                &SecureBootEnable
+                );
+
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Cannot read SecureBootEnable variable: %r\n", Status));
+    /* Get the state from the from data if we failed to read the variable */
+    SecureBootEnable = ConfigData->AttemptSecureBoot;
```
krystian-hebel, line 4188: Why not read from PCD if there is no variable? That way `ConfigData` wouldn't have to be passed to this function.

On a tangent note, as I can't add comments to unmodified code, I think the [test done in lines 3227-3229](https://github.com/Dasharo/edk2/blob/21462205e5e95e51f8f6b00da3c43e85da9c9cdf/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c#L3227-L3229) should be moved outside of `if ((*SetupMode) == USER_MODE)`. Right now the `Enable Secure Boot` checkbox isn't checked when `SecureBootEnable` is true, but there is no PK provisioned. There already is a [proper message about PK](https://github.com/Dasharo/edk2/blob/21462205e5e95e51f8f6b00da3c43e85da9c9cdf/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigStrings.uni#L21) printed in that case, I think having this option grayed out but checked would be less confusing than having it empty.

BTW is it normal that `grayoutif` is bold instead of gray on serial? It doesn't intuitively suggest that the option is disabled...

krystian-hebel (changes requested):
From technical point of view it works more or less as discussed, but UX is poor due to `Enable Secure Boot         [ ]` when `SecureBootEnable` is true, but no PK.

However, I don't think this by itself meets the expectations described by the vendor. Enrolling PK from OS **does not** enable SB without setting `SecureBootEnable` in setup menu, at least with current `CONFIG_EDK2_SECURE_BOOT_DEFAULT_ENABLE` state. It is acceptable, but:
- Should be documented (https://docs.dasharo.com -> guides?)
  - preferably with instructions for extracting current/default KEK, db and dbx, and signing them with new PK to not have to sign existing bootloaders,
  - similar instructions for users that want to build edk2 with custom keys, maybe?
- It would be easier for us to have the same default across all platforms.
- Currently, SMMSTORE region is cleared with each update, which makes whole custom SB configuration tedious, but it may be fixed in the future with the help of capsules (in the wider meaning, not necessarily capsule updates).


Also, please add commit messages.

SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
miczyg1: > Why not read from PCD if there is no variable? That way ConfigData wouldn't have to be passed to this function.

This is not about restoring the default, but preserving current state. We want the reset keys to not change SB state.

miczyg1: > On a tangent note, as I can't add comments to unmodified code, I think the [test done in lines 3227-3229](https://github.com/Dasharo/edk2/blob/21462205e5e95e51f8f6b00da3c43e85da9c9cdf/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c#L3227-L3229) should be moved outside of if ((*SetupMode) == USER_MODE). Right now the Enable Secure Boot checkbox isn't checked when SecureBootEnable is true, but there is no PK provisioned. There already is a [proper message about PK](https://github.com/Dasharo/edk2/blob/21462205e5e95e51f8f6b00da3c43e85da9c9cdf/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigStrings.uni#L21) printed in that case, I think having this option grayed out but checked would be less confusing than having it empty.

Checked and grayed out even when PK is not provisioned? Is such a state even valid? PK must be provisoned in order to enable Secure Boot. Also being unable to disable Secure Boot like that is a bad UX IMO.

> BTW is it normal that grayoutif is bold instead of gray on serial? It doesn't intuitively suggest that the option is disabled...

:shrug:

pietrushnic:
@macpijan @BeataZdunczyk I'm not sure what is the context of these changes, but they seem to be closely related to planned Arch4221: UEFI Secure Boot so if you could link that effort internally, it would be great.

krystian-hebel (approved):

PR #147, opened 2024-06-30
Enable capsule updates

SergiiDmytruk:
This lacks FMP for update to actually do anything.

***

New CAPSULE_ENABLE define in DasharoPayloadPkg.dsc adds required libraries/DXEs/PCDs and enables building of CapsuleApp which is useful for testing.

BlSupportPei now switches boot mode to BOOT_ON_FLASH_UPDATE if there are any capsules.  It's not guarded by any PCD under the assumption that coreboot won't pass capsules if EDK shouldn't be handling them.

EsrtDxe is enabled to manage and cache ESRT entries (caching is supposedly necessary after a capsule is processed, although don't really understand why).  ProcessCapsules() internally looks up EsrtManagementProtocol and calls SyncEsrtFmp() to import data from available FMP instances.

PlatformBootManagerLib was made to call ProcessCapsules() (twice).

PR #154, opened 2024-07-24
Implement capsule flashing via extended SMMSTORE v2

SergiiDmytruk:
Due to https://github.com/Dasharo/dasharo-issues/issues/800#issuecomment-2247928107, update capsules aren't accepted unless ME is HAP-disabled.  This has a side effect of preventing update in Q35.  Don't yet know how to test for QEMU or otherwise circumvent the check for Q35, but that's a minor thing.

***

- FUM is enabled before rebooting to handle a persistent capsule
- SmmStoreLib can now be used in `DXE` modules, not just in `DXE_RUNTIME`, and is used by FmpDxe which gets embedded into a capsule, the embedding is done by adding a section like this to JSON file passed to `GenerateCapsule`:
```json
    "EmbeddedDrivers": [
        {
            "Driver": "../Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/FmpDxe.efi"
        }
    ],
```

### DIFF
--- a/DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
+++ b/DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
@@ -439,6 +439,64 @@ DasharoEnableFUM (
   return Status;
 }
 
+/**
+  Check whether capsule updates which survive a warm system reset are permitted
+  by current configuration.
+
+  @retval TRUE   Persistent capsules can be accepted by UpdateCapsule().
+  @retval FALSE  UpdateCapsule() must fail with an error for such a capsule.
+**/
+BOOLEAN
+EFIAPI
+DasharoCapsulesCanPersistAcrossReset (
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+  UINT8       MeMode;
+  UINTN       VarSize;
+
+  //
+  // Assuming there is no ME if the corresponding menu is not enabled.
+  //
+  if (!FixedPcdGetBool (PcdShowIntelMeMenu)) {
+    return TRUE;
+  }
+
+  //
+  // Guard against variable's value obviously lying about the state.
+  //
+  if (!FixedPcdGetBool (PcdIntelMeHapAvailable)) {
+    return FALSE;
+  }
+
+  MeMode = DASHARO_ME_MODE_ENABLE;
+  VarSize = sizeof (MeMode);
+
+  Status = gRT->GetVariable (
+      DASHARO_VAR_ME_MODE,
+      &gDasharoSystemFeaturesGuid,
+      NULL,
+      &VarSize,
+      &MeMode
+      );
+  ASSERT_EFI_ERROR (Status);
+
+  //
+  // HAP-disabled ME doesn't do anything, including writing to system flash,
+  // which is what we need for a firmware update that relies on a warm reset.
+  // coreboot assumes that HECI/soft-disabled state of ME isn't as good as
+  // HMRFPO and switches to HMRFPO doing a global reset which loses in-RAM
+  // capsules.
+  //
+  // Checking variable's value should be enough, if somebody manually set it to
+  // an invalid value, the update there will be a reboot without a capsule
+  // update.  A more reliable solution would be to pass this information from
+  // coreboot.
+  //
+  return MeMode == DASHARO_ME_MODE_DISABLE_HAP;
+}
+
 EFI_STATUS
 EFIAPI
 DasharoVariablesLibConstructor (
--- a/DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
+++ b/DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
@@ -413,6 +413,32 @@ DasharoMeasureVariables (
   return Status;
 }
 
+/**
+  Enable firmware update mode (FUM) for the next boot.
+
+  @retval RETURN_SUCCESS  FUM was successfully enabled.
+**/
+EFI_STATUS
+EFIAPI
+DasharoEnableFUM (
+  VOID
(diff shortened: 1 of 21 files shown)


DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
```diff
@@ -413,6 +413,90 @@ DasharoMeasureVariables (
...
+  )
+{
+  EFI_STATUS  Status;
+  UINT8       MeMode;
+  UINTN       VarSize;
+
+  //
+  // Assuming there is no ME if the corresponding menu is not enabled.
+  //
+  if (!FixedPcdGetBool (PcdShowIntelMeMenu)) {
```
krystian-hebel, line 462: Shouldn't this make base Q35 work? It is [selected when building with full menu](https://github.com/Dasharo/coreboot/blob/dasharo/configs/config.emulation_qemu_x86_q35_uefi_all_menus#L37), but I have `# CONFIG_EDK2_DASHARO_INTEL_ME_OPTIONS is not set` in `.config` made from https://github.com/Dasharo/coreboot/blob/dasharo/configs/config.emulation_qemu_x86_q35_uefi. This may require changes to our CI, but I don't see why capsule wouldn't work with QEMU.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -0,0 +1,1089 @@
...
+typedef struct {
+  EFI_GUID  FwGuid;
+  UINT32    FwVersion;
+  UINT32    FwLsv;
+  UINT32    FwSize;
+} FwInfo;
+
+/**
+  This function requests firmware information on the first call, caches it and
+  return on all calls afterwards.
```
krystian-hebel, line 34:
```suggestion
  returns on all calls afterwards.
```

MdeModulePkg/Universal/CapsuleRuntimeDxe/CapsuleReset.c
```diff
@@ -24,5 +24,9 @@ IsPersistAcrossResetCapsuleSupported (
   VOID
   )
 {
+  if (!DasharoCapsulesCanPersistAcrossReset()) {
+    return FALSE;
+  }
```
krystian-hebel, line 29: Can `FeaturePcdSet()` be used instead of adding this check here?

DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
SergiiDmytruk: Capsules do work in `config.emulation_qemu_x86_q35_uefi`, but if `config.emulation_qemu_x86_q35_uefi_all_menus` is to be used for firmware tests I thought the capsules should work there too.

MdeModulePkg/Universal/CapsuleRuntimeDxe/CapsuleReset.c
SergiiDmytruk: There is no such function, feature flags are fixed at build time.  `AutoGen.h` contains:
```c
//#define _PCD_SET_MODE_BOOL_PcdSupportUpdateCapsuleReset  ASSERT(FALSE)  // It is not allowed to set value for a FIXED_AT_BUILD PCD
```

krystian-hebel: Right, it did look strange to me, but such function [is mentioned in the edk2 docs](https://github.com/tianocore-docs/edk2-ModuleWriteGuide/blob/master/3_module_development/32_creating_a_module.md#table-5-pcd-access-functions). I guess we have to live with it, maybe during upstreaming someone will propose another solution.

DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
krystian-hebel: We don't have this in the scope of project, and even though it's much easier to test in QEMU than on hardware, we never promised such tests.

That said, it is possible that we will have to make edk2 aware of it running in QEMU to resolve https://github.com/Dasharo/dasharo-issues/issues/898, due to different `gEfiMdePkgTokenSpaceGuid.PcdFSBClock`. The same flag could be used here, but that is something for the future.

krystian-hebel (approved):

DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
SergiiDmytruk: I was thinking of adding `EDK2_APU_CONFIG` and `EDK2_INTEL_ME_HAP` to allow forcing those on in `config.emulation_qemu_x86_q35_uefi_all_menus`.  QEMU config is for coreboot and this would allow to keep related changes in coreboot.  `PcdFSBClock` can be handled similarly.

MdeModulePkg/Universal/CapsuleRuntimeDxe/CapsuleReset.c
SergiiDmytruk: Upstream shouldn't see this commit because it depends on many other Dasharo-specific modifications.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
SergiiDmytruk: I was waiting for your replies before pushing the typo fix.  Force-pushed https://github.com/Dasharo/edk2/pull/147 to address this.

PR #161, opened 2024-08-09
DasharoPayloadPkg: Fix PciHostBridgeLib for resources above 4G

mkopec:
This is a port of upstream commits c248802e40dc2cea007a7063b73feb0d04809261 and bfefdc2c49ca9487b7aa0df196b2aca6c0c170a2

On modern platforms with TBT devices the coreboot resource allocator opens large PCI bridge MMIO windows above 4GiB to place hotplugable PCI BARs there as they won't fit below 4GiB. In addition modern GPGPU devices have very big PCI bars that doesn't fit below 4GiB.

The PciHostBridgeLib made lots of assumptions about the coreboot resource allocator that were not verified at runtime and are no longer true.

Remove all of the 'coreboot specific' code and implement the same logic as OvmfPkg's ScanForRootBridges.

Fixes assertion
"ASSERT [PciHostBridgeDxe] Bridge->Mem.Limit < 0x0000000100000000ULL".

Also fixes PCI enumeration on Clevo V560TND with discrete graphics whose resources are allocated above 4G

### DIFF
--- a/DasharoPayloadPkg/Library/PciHostBridgeLib/PciHostBridgeSupport.c
+++ b/DasharoPayloadPkg/Library/PciHostBridgeLib/PciHostBridgeSupport.c
@@ -20,88 +20,6 @@
 #include "PciHostBridge.h"
 
 /**
-  Adjust the collected PCI resource.
-
-  @param[in]  Io               IO aperture.
-
-  @param[in]  Mem              MMIO aperture.
-
-  @param[in]  MemAbove4G       MMIO aperture above 4G.
-
-  @param[in]  PMem             Prefetchable MMIO aperture.
-
-  @param[in]  PMemAbove4G      Prefetchable MMIO aperture above 4G.
-**/
-VOID
-AdjustRootBridgeResource (
-  IN  PCI_ROOT_BRIDGE_APERTURE *Io,
-  IN  PCI_ROOT_BRIDGE_APERTURE *Mem,
-  IN  PCI_ROOT_BRIDGE_APERTURE *MemAbove4G,
-  IN  PCI_ROOT_BRIDGE_APERTURE *PMem,
-  IN  PCI_ROOT_BRIDGE_APERTURE *PMemAbove4G
-)
-{
-  UINT64  Mask;
-
-  //
-  // For now try to downgrade everything into MEM32 since
-  // - coreboot does not assign resource above 4GB
-  // - coreboot might allocate interleaved MEM32 and PMEM32 resource
-  //   in some cases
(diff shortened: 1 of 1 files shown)


mkopec:
Likely to resolve GPU and Thunderbolt issues on other platforms, too.

SergiiDmytruk:
https://github.com/tianocore/edk2/commit/991c9787d37c0faac540ebe2d8497647a7ff0606 doesn't seem to be in upstream.  Looks like the correct commit version is https://github.com/tianocore/edk2/commit/c248802e40dc2cea007a7063b73feb0d04809261.  https://github.com/tianocore/edk2/commit/bfefdc2c49ca9487b7aa0df196b2aca6c0c170a2 also seems to be included here, but not mentioned.

mkopec:
You're right, I've updated the PR message.

SergiiDmytruk:
GitHub used commit's message for PR's message and I thought you would update the commit.

mkopec:
Updated commit message as well @SergiiDmytruk

miczyg1:
@mkopec with top down resource allocation the V560TU does boot in recovery saying that CSE Lite SKU firmware failure occurred (failed to send EOP). Kind of expected that coreboot cannot send EOP if HECI BAR is 64 bit. When I soft disabled ME, the recovery was gone as expected. The GOP x64 EFI also worked well with 64bit BAR so I think we may consider it working.

EDIT: touchpad not working though...

mkopec:
@miczyg1 yeah there's a bug with the designware i2c driver in coreboot when topdown resource alloc is enabled :\

miczyg1:
> @miczyg1 yeah there's a bug with the designware i2c driver in coreboot when topdown resource alloc is enabled :\

There is even more. The function for getting GFX BAR for the framebuffer can only pass 32bit uintptr_t (in case the GOP is used), so of course the LB_TAG_FRAMBUFFER has a truncated address and EDK2 GraphicsOutputDxe can't find the BAR matching the address :)

mkopec:
@miczyg1 @SergiiDmytruk @krystian-hebel ping :) We've been using this change for a while in novacustom MTL release candidates with dGPU support

SergiiDmytruk:
I have a very vague idea of what the code is doing, hence can't approve.  There are no details about now invalid coreboot assumptions, why they are invalid and what the new implementation is doing differently in those cases.  If nobody has enough insight, then we can go by the testing results and upstream's review (like we've taken upstream commits from 4 years during rebase).

miczyg1 (approved):

PR #163, opened 2024-08-15
Fix MbedTLS allocations at runtime

SergiiDmytruk:
As was recently reported on Matrix for CachyOS on Nova Custom V56 and in https://github.com/Dasharo/dasharo-issues/issues/1001 for FreeBSD on Protectli V1410, EFI runtime can crash when dealing with SecureBoot EFI variables as in https://github.com/Dasharo/edk2/pull/129#issuecomment-2150198582.

I originally fixed it only in a single file which showed up in https://github.com/Dasharo/edk2/pull/129#issuecomment-2153493525 and somehow concluded that it should be enough which wasn't true.  This attempt should be more successful, making it a PR so that more people can take a look and/or test.

The issue can be reproduced in QEMU and probably anywhere else.  I installed Debian, reset SecureBoot keys, downloaded https://github.com/Foxboron/sbctl/releases/tag/0.15.4 binary and tried following https://wiki.cachyos.org/configuration/secure_boot_setup/#setting-up-sbctl.  `sbctl enroll-keys -m` failed the same way it was described on Matrix.  The build from this branch succeeded.

Read commit messages for more details.

P.S. Maybe this should also be sent upstream, which lacks these and some other fixes made previously.

### DIFF
--- a/CryptoPkg/Library/BaseCryptLib/SysCall/TimerWrapper.c
+++ b/CryptoPkg/Library/BaseCryptLib/SysCall/TimerWrapper.c
@@ -135,7 +135,8 @@ gmtime (
   const time_t  *timer
   )
 {
-  struct tm  *GmTime;
+  STATIC struct tm  GmTime;
+
   UINT64     DayNo;
   UINT64     DayRemainder;
   time_t     Year;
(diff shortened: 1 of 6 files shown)


krystian-hebel (approved):

macpijan:
> P.S. Maybe this should also be sent upstream, which lacks these and some other fixes made previously.

That's great idea, otherwise we will keep maintaining more and more diff in the long run.

PR #164, opened 2024-08-27
OvmfPkg: DxeTcg2PhysicalPresenceLib: fix changing of PCR banks

arturkow2:
Tcg2PhysicalPresenceLibSubmitRequestToPreOSFunction writes to TPM2 physical presence PPI provided by coreboot (a memory region preserved across reboots). CPU caches must be explicitly flushed prior to platform reboot or request written to PPI will be lost.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
         fi
 
     - name: Upload Artifacts
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
       with:
         name: ovmf-artifacts  # Name for the artifact
         path: |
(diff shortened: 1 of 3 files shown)


miczyg1:
@arturkow2 please do not open PRs based outdated history of target branch... Be sure to pull most recent code before yo ubranch out for development. I already rebased this branch on dasharo.

miczyg1:
I have tested the changes on VP6650 with Infineon dTPM and it didn't work... I was not prompted to confirm the SHA bank change before entering boot menu or setup.

EDIT: Wait, this platform is actually the worst possible piece to test this on. This platform can't reset properly, as it always goes through power cycle...

miczyg1:
Ok it works, but once I apply new PCR banks settings and reset the platform I am stuck on the prompt to change PCR banks forever:

```
A configuration change was requested to change the boot measurements to use PCR 
bank(s) of this computer's TPM (Trusted Platform Module)

WARNING: Changing the PCR bank(s) of the boot measurements may prevent the Opera
ting System from properly processing the measurements. Please check if your Oper
ating System supports the new PCR bank(s).

WARNING: Secrets in the TPM that are bound to the boot state of your machine may
 become unusable.

Current PCRBanks is 0x2. (SHA256)
New PCRBanks is 0x2. (SHA256)
 
Press F12 to change the boot measurements to use PCR bank(s) of the TPM 
Press ESC to reject this change request and continue
```

I have to press ESC to exit the loop. It looks like the request is not cleared after it is serviced. We have to clear it so that it does not persist in RAM and is handled on each boot... This behavior was observed on VP4670

miczyg1:
Ok, now it works when the request is cleared and flushed

miczyg1 (approved):

PR #166, opened 2024-09-19
Preserve user data during firmware updates

SergiiDmytruk:
Issue: https://github.com/Dasharo/dasharo-issues/issues/809

Lots of code changes, but most of that is imported code:
 - FMAP from coreboot (updated to be more EDK-like)
 - CBFS from `cbfstool` in coreboot (essentially unchanged)
 - SMMSTORE code from `smmstoretool` in coreboot (reduced to bare minimum)
 - code for iterating EFI variables is from `DasharoVariablesLib` (adjusted to fit the use case)

The main code is in `FmpDeviceSmmLib` in a new file `Flashing.c`.

### DIFF
--- a/DasharoPayloadPkg/Library/CbfsLib/adapt.h
+++ b/DasharoPayloadPkg/Library/CbfsLib/adapt.h
@@ -1,0 +1,39 @@
+#ifndef ADAPT_H__
+#define ADAPT_H__
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/MemoryAllocationLib.h>
+
+typedef UINTN size_t;
+typedef INTN ptrdiff_t;
+typedef UINT8 uint8_t;
+typedef UINT16 uint16_t;
+typedef UINT32 uint32_t;
+typedef UINT64 uint64_t;
+typedef INT8 int8_t;
+typedef INT16 int16_t;
+typedef INT32 int32_t;
+typedef INT64 int64_t;
+typedef BOOLEAN bool;
+
+#define assert(x) ASSERT (x)
+#define memcmp(a, b, l) CompareMem ((a), (b), (l))
+#define memcpy(a, b, l) CopyMem ((a), (b), (l))
+#define memmove(a, b, l) CopyMem ((a), (b), (l))
+#define memset(b, v, l) SetMem ((b), (l), (v))
+#define strcpy(d, s) AsciiStrCpyS ((d), AsciiStrSize (s), (s))
+#define strlen(s) AsciiStrLen (s)
+#define strcasecmp(s, t) AsciiStriCmp ((s), (t))
+#define malloc(s) AllocatePool (s)
+#define free(p) FreePool (p)
+
+#define false FALSE
+#define true TRUE
+
+#define ERROR(...) DEBUG ((DEBUG_ERROR, __VA_ARGS__))
+#define INFO(...) DEBUG ((DEBUG_INFO, __VA_ARGS__))
+#define WARN(...) DEBUG ((DEBUG_WARN, __VA_ARGS__))
+
+#endif /* ADAPT_H__ */
--- a/DasharoPayloadPkg/Library/EfiVarsLib/EfiVarsLib.c
+++ b/DasharoPayloadPkg/Library/EfiVarsLib/EfiVarsLib.c
@@ -1,0 +1,246 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent */
+
+#include <Library/EfiVarsLib.h>
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/SmmStoreLib.h>
+
+// Firmware volume is what's stored in SMMSTORE region of CBFS.  It wraps
+// variable store.
+//
+// Variable store is part of firmware volume.  This unit doesn't deal with its
+// header only with data that follows.
+
+STATIC
+UINT16
+CalcChecksum (
+  CONST UINT16 *Hdr,
+  UINTN Size
+  )
+{
+  UINT16  Checksum;
+  UINTN   Idx;
+
+  ASSERT (Size % 2 == 0 && "Header can't have odd length.");
+
+  Checksum = 0;
+  for (Idx = 0; Idx < Size / 2; ++Idx)
+    Checksum += Hdr[Idx];
+
+  return Checksum;
+}
+
+STATIC
+BOOLEAN
+InitFv (
+  IN OUT MemRange  Fv
+  )
+{
+  EFI_STATUS                  Status;
+  UINTN                       BlockSize;
+  UINTN                       VolHdrLen;
+  EFI_FIRMWARE_VOLUME_HEADER  *VolHdr;
+  VARIABLE_STORE_HEADER       *VarStoreHdr;
+
+  Status = SmmStoreLibGetBlockSize (&BlockSize);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a(): SmmStoreLibGetBlockSize() failed with: %r\n",
+      __FUNCTION__,
+      Status
+      ));
+    return FALSE;
+  }
+
+  if (Fv.Length % BlockSize != 0) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a(): firmware Volume size is not a multiple of 64KiB\n",
+      __FUNCTION__
+      ));
+    return FALSE;
+  }
+
+  SetMem (Fv.Start, Fv.Length, 0xff);
+
+  VolHdrLen = sizeof (*VolHdr) + sizeof (EFI_FV_BLOCK_MAP_ENTRY);
+
+  VolHdr = (EFI_FIRMWARE_VOLUME_HEADER *) Fv.Start;
+  SetMem (VolHdr, sizeof (*VolHdr), 0x00);
+
+  VolHdr->FileSystemGuid = gEfiSystemNvDataFvGuid;
+  VolHdr->FvLength       = Fv.Length;
+  VolHdr->Signature      = EFI_FVH_SIGNATURE;
+  VolHdr->Attributes     = EFI_FVB2_READ_ENABLED_CAP
+                         | EFI_FVB2_READ_STATUS
+                         | EFI_FVB2_WRITE_ENABLED_CAP
+                         | EFI_FVB2_WRITE_STATUS
+                         | EFI_FVB2_STICKY_WRITE
+                         | EFI_FVB2_MEMORY_MAPPED
+                         | EFI_FVB2_ERASE_POLARITY;
+  VolHdr->HeaderLength   = VolHdrLen;
+  VolHdr->Revision       = EFI_FVH_REVISION;
+
+  VolHdr->BlockMap[0].NumBlocks = Fv.Length / BlockSize;
+  VolHdr->BlockMap[0].Length    = BlockSize;
+  VolHdr->BlockMap[1].NumBlocks = 0;
+  VolHdr->BlockMap[1].Length    = 0;
+
+  VolHdr->Checksum = ~CalcChecksum ((CONST UINT16 *) VolHdr, VolHdrLen) + 1;
+
+  VarStoreHdr = (VARIABLE_STORE_HEADER *) (Fv.Start + VolHdrLen);
+  SetMem (VarStoreHdr, sizeof (*VarStoreHdr), 0x00);
+
+  // Authentication-related fields will be filled with 0xff.
+  VarStoreHdr->Signature = gEfiAuthenticatedVariableGuid;
+  // Actual size of the storage is block size, the rest is
+  // Fault Tolerant Write (FTW) space and the FTW spare space.
+  VarStoreHdr->Size      = BlockSize - VolHdrLen;
+  VarStoreHdr->Format    = VARIABLE_STORE_FORMATTED;
+  VarStoreHdr->State     = VARIABLE_STORE_HEALTHY;
+
+  return TRUE;
+}
+
+BOOLEAN
+EFIAPI
+EfiVarsInit (
+  IN OUT MemRange  Fv,
+  OUT EfiVars      *Storage
+  )
+{
+  CONST EFI_FIRMWARE_VOLUME_HEADER  *VolHdr;
+  CONST VARIABLE_STORE_HEADER       *VarStoreHdr;
+  UINT8                             *VolData;
+  UINTN                             VolDataSize;
+
+  if (!InitFv (Fv)) {
+    DEBUG ((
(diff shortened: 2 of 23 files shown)


DasharoPayloadPkg/Library/FmapLib/FmapLib.c
```diff
@@ -0,0 +1,199 @@
...
+  IN CHAR8  Chr
+  )
+{
+  return (Chr > 32 && Chr < 126);
+}
+
+/* Make a best-effort assessment if the given fmap is real */
+STATIC
+UINTN
+is_valid_fmap (
```
krystian-hebel: All other function names are CamelCased, this should be too.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -956,9 +958,13 @@ FmpDeviceSetImageWithStatus (
     return Status;
   }
 
-  ReadBuffer = AllocatePool (BlockSize);
-  if (ReadBuffer == NULL) {
-    DEBUG ((DEBUG_ERROR, "%a(): failed to allocate read buffer\n", __FUNCTION__));
+  CurrentImage = ReadCurrentFirmware ();
```
krystian-hebel, line 977: How long does the reading process take? If it is noticeable maybe the progress bar should be moved by few percents (before and after reading) to calm the users.

DasharoPayloadPkg/Library/FmapLib/FmapLib.c
```diff
@@ -0,0 +1,199 @@
...
+  IN CONST UINT8 *Image,
+  IN UINTN Len
+  )
+{
+  INTN Offset;
+  BOOLEAN FmapFound;
+
+  FmapFound = FALSE;
+
+  for (Offset = 0; Offset < Len - AsciiStrLen (FMAP_SIGNATURE); Offset++) {
```
krystian-hebel: `is_valid_fmap` accesses fields after the signature, this would be a potential out of bounds read. Is this code copied from coreboot/cbfstool? IIRC coreboot uses defines for location and size of FMAP (which obviously isn't possible here), but cbfstool may do things differently, after all it can read any ROM, not just the one built the same layout.

```suggestion
  for (Offset = 0; Offset < Len - sizeof (Fmap); Offset++) {
```

DasharoPayloadPkg/Library/FmapLib/FmapLib.c
```diff
@@ -0,0 +1,199 @@
...
+   * and then decrease the Stride on each iteration. Also, check for a
+   * remainder when modding the Offset with the previous Stride. This
+   * makes it so that each Offset is only checked once.
+   */
+  for (Stride = Len / 2; Stride >= 16; Stride /= 2) {
+    if (FmapFound)
+      break;
+
+    for (Offset = 0;
+         Offset < Len - AsciiStrLen (FMAP_SIGNATURE);
```
krystian-hebel:
```suggestion
         Offset < Len - sizeof (Fmap);
```

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -81,6 +143,82 @@ ReadCurrentFirmware (
...
+  if (CurrentRegion->size != UpdatedRegion->size) {
+    DEBUG ((
+      DEBUG_WARN,
+      "%a(): %a regions don't match in size (current: 0x%x, updated: 0x%x)\n",
+      __FUNCTION__,
+      RegionName,
+      CurrentRegion->size,
+      UpdatedRegion->size
+      ));
+    return REGION_OF_DIFFERENT_SIZE;
```
krystian-hebel, line 215: Does it make sense to copy the region anyway if current size is smaller than updated?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -184,6 +187,18 @@ MigrateRegion (
     return REGION_NOT_IN_DST;
   }
 
+  if (OffsetSensitive && CurrentRegion->offset != UpdatedRegion->offset) {
+    DEBUG ((
+      DEBUG_WARN,
+      "%a(): %a regions don't match in size (current: 0x%x, updated: 0x%x)\n",
+      __FUNCTION__,
+      CurrentRegion->size,
+      UpdatedRegion->size,
```
krystian-hebel: Copy-paste error, this should mention offsets, not sizes.

DasharoPayloadPkg/Library/CbfsLib/adapt.h
```diff
@@ -0,0 +1,39 @@
...
+typedef INT32 int32_t;
+typedef INT64 int64_t;
+typedef BOOLEAN bool;
+
+#define assert(x) ASSERT (x)
+#define memcmp(a, b, l) CompareMem ((a), (b), (l))
+#define memcpy(a, b, l) CopyMem ((a), (b), (l))
+#define memmove(a, b, l) CopyMem ((a), (b), (l))
+#define memset(b, v, l) SetMem ((b), (l), (v))
+#define strcpy(d, s) AsciiStrCpyS ((d), strlen (s) + 1, (s))
```
krystian-hebel: Not sure if you're aware, but there is actually `AsciiStrSize()` that does `strlen (s) + 1`. I'm fine with keeping it in current form, though, it is more understandable to someone familiar with standard C than `AsciiStrSize()`. Feel free to mark this as resolved if you don't want to change it.

MdeModulePkg/Logo/Logo.c
```diff
@@ -95,20 +95,19 @@ GetImage (
...
-  } else {
-    // No logo in CBMEM, fallback to builtin
-    *Attribute = mLogos[Current].Attribute;
-    *OffsetX   = mLogos[Current].OffsetX;
-    *OffsetY   = mLogos[Current].OffsetY;
-    return mHiiImageEx->GetImageEx (mHiiImageEx, mHiiHandle, mLogos[Current].ImageId, Image);
+    DEBUG ((DEBUG_ERROR, "Failed to translate logo image: %r\n", Status));
   }
+
+  // No or bad logo in CBMEM, fallback to builtin
```
krystian-hebel, line 106: Ah, yes, the dependency of progress bar on logo...

It goes beyond just the existence of logo, its dimensions are used to determine size and position of progress bar. In https://github.com/Dasharo/edk2/pull/167 I'm implementing a workaround to have it consistent regardless of logo dimensions, but as it is a separate driver, it is good to have a fallback here.

This is just informative, you can resolve this comment after reading.

DasharoPayloadPkg/Library/EfiVarsLib/EfiVarsLib.c
```diff
@@ -0,0 +1,251 @@
...
+  EfiVar *Var;
+
+  new_var = AllocatePool (sizeof (*new_var));
+  if (new_var == NULL)
+    return NULL;
+
+  SetMem (new_var, sizeof (*new_var), 0);
+  new_var->Attrs = EFI_VARIABLE_NON_VOLATILE
+                 | EFI_VARIABLE_BOOTSERVICE_ACCESS
+                 | EFI_VARIABLE_RUNTIME_ACCESS;
```
krystian-hebel: Unless I'm missing something, this is unnecessary. The caller writes proper attributes, and they may be different than these.

krystian-hebel (commented):
The last commit has a side effect of consolidating the variable store, it may be useful (unless SMMSTOREv2 already uses garbage collection, I know v1 didn't).

DasharoPayloadPkg/Library/FmapLib/FmapLib.c
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-c9213a732aef6113cdcf417b51b73bea7dee0cd1ae2810f90c794db481ba9762

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
SergiiDmytruk: About 10 seconds.  Took a while to figure this out because I assumed things to work on MSI, but this required trying it out there and nothing worked.  Previously, the reads could fail silently and they actually did.

Turned out that first 4 MiB of flash aren't mmapped, only "bios" region is.  Not sure how to resolve this properly, locally I've added `4 * MiB` to size at https://github.com/Dasharo/coreboot/blob/bd294ab1d22b4010fafa0120e2eda26964316096/src/soc/intel/common/block/fast_spi/mmap_boot.c#L123 which made the whole flash available to coreboot.  I guess this only affects reads as erases or writes didn't fail.

Made this read to move progress by about 5%: https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307f

DasharoPayloadPkg/Library/FmapLib/FmapLib.c
SergiiDmytruk: > `is_valid_fmap` accesses fields after the signature, this would be a potential out of bounds read.

Right.  I didn't notice it.  Other than stylistic changes, I've only done adjustments necessary to fix compilation.

> Is this code copied from coreboot/cbfstool?

Yes, https://github.com/coreboot/coreboot/blob/cb7c58f6266b10fdd1ffc3309aa929b11a81f6c5/util/cbfstool/flashmap/fmap.c#L80

https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-c9213a732aef6113cdcf417b51b73bea7dee0cd1ae2810f90c794db481ba9762R94

DasharoPayloadPkg/Library/FmapLib/FmapLib.c
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-c9213a732aef6113cdcf417b51b73bea7dee0cd1ae2810f90c794db481ba9762R134

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
SergiiDmytruk: This really depends on the contents and what updated code expects to find in the region, because it can rely on the region size and try processing data after what was copied.  I wouldn't do this for `ROMHOLE`. `BOOTSPLASH` could be changed to move a CBFS file instead of copying the region, but I didn't want to bother due the logo being compressed and that needs special handling.

`cbfs_image.c` is actually trimmed down. I forgot that I took the files from https://github.com/pcengines/cb-order/tree/master/third-party where I've already dealt with includes and such before.  Compared against version on coreboot's master and imported improvements: https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-12229f96e4e801a7525d42f6e6437d39cc663b19f4944597ca29669092b8c57d

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
SergiiDmytruk: Not only that: `RegionName` should precede offsets.

https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-df48c35c52e8b3221b927ccc9e290725c37320ea2077f5b0862609923f3a8aa0

DasharoPayloadPkg/Library/CbfsLib/adapt.h
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-1aa89257ae0456694c7a1113e3c13d1c12cbc9375e27ac46a9c5a95d36947abd

Updated, I just didn't have a preference one way or the other.

DasharoPayloadPkg/Library/EfiVarsLib/EfiVarsLib.c
SergiiDmytruk: Yes, it's a leftover from original code.

https://github.com/Dasharo/edk2/compare/ab54493809413c1639ff172e219168385a576c97..a53ef1c4ebc4be7222fa234b1ce6cb61e6e4a3ee#diff-c8b499485b05281f2f4aefefa1ca676cdb9b81f8a5fc2cae271d535287619edf

SergiiDmytruk:
> The last commit has a side effect of consolidating the variable store, it may be useful (unless SMMSTOREv2 already uses garbage collection, I know v1 didn't).

SMMSTORE is a storage protocol not aware of what it's storing.  I think coreboot's code can only grow SMMSTORE, but EDK has https://github.com/Dasharo/edk2/blob/250710a33772c74376a321471704ddaf09e70d3c/MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c#L496 which seems to perform cleanup at least on some updates of variables.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
krystian-hebel: > Made this read to move progress by about 5%

That part is resolved.

> Turned out that first 4 MiB of flash aren't mmapped, only "bios" region is. Not sure how to resolve this properly, locally I've added 4 * MiB to size at https://github.com/Dasharo/coreboot/blob/bd294ab1d22b4010fafa0120e2eda26964316096/src/soc/intel/common/block/fast_spi/mmap_boot.c#L123 which made the whole flash available to coreboot.

Interesting, especially since it then creates a device for whole `CONFIG_ROM_SIZE` in https://github.com/Dasharo/coreboot/blame/bd294ab1d22b4010fafa0120e2eda26964316096/src/soc/intel/common/block/fast_spi/mmap_boot.c#L134. There is no discussion as to size on gerrit, so perhaps this mismatch wasn't deliberate, or maybe it was done so that SPI addresses are correct.

I think the best option would be to add another Kconfig for mmaping whole ROM instead of just BIOS region, and make it selected when capsules are enabled.

krystian-hebel: Anyway, this is coreboot issue, it doesn't block merging this PR.

krystian-hebel (approved):

PR #168, opened 2024-09-24
DasharoPayloadPkg: Use INIT-SIPI-SIPI for first AP wakeup

miczyg1:
It has been observed on MinnowBoard Turbot that the detected CPU count is lower than the number of all cores. In the tested unit, CPU is dual core, so only the BSP is detected. However, similar situation is observed on MTL laptop, where a total of 22 cores should be reported, but only 1 is detected.

After EDK2 rebase a new PCD has been added, PcdFirstTimeWakeUpAPsBySipi which, when enabled (by default), sends only SIPI to APs. When disabled sends full INIT-SIPI-SIPI sequence. Only the latter case causes all the APs to wake up and be detected.

### DIFF
--- a/DasharoPayloadPkg/DasharoPayloadPkg.dsc
+++ b/DasharoPayloadPkg/DasharoPayloadPkg.dsc
@@ -540,6 +540,7 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciSerialParameters|$(PCI_SERIAL_PARAMETERS)
 
   gUefiCpuPkgTokenSpaceGuid.PcdCpuMaxLogicalProcessorNumber|$(MAX_LOGICAL_PROCESSORS)
+  gUefiCpuPkgTokenSpaceGuid.PcdFirstTimeWakeUpAPsBySipi|FALSE
 
 !if $(PS2_KEYBOARD_ENABLE) == TRUE
   gDasharoPayloadPkgTokenSpaceGuid.PcdSkipPs2Detect|FALSE


krystian-hebel:
https://github.com/Dasharo/dasharo-issues/issues/1054

krystian-hebel:
While this may fix the visible part of the issue, it may be worth to think what is causing it. CPU may ignore SIPI if it already is executing an interrupt of the same or higher priority, but lower than INIT. SIPI would be latched and executed afterwards, unless the interrupt handler never finishes.

There are three priority classes that can do this: maskable hardware interrupts (IIRC this includes SIPI), NMI and traps on previous instruction, like breakpoints or debug trap exceptions (i.e. `DEBUG_NULL_DEREF_BREAKPOINTS` may be involved).

miczyg1:
> While this may fix the visible part of the issue, it may be worth to think what is causing it.

Caused by the rebase of EDK2. New changes were added to the MpLib. coreboot leaves the APs in a halt loop, so they won't respond to SIPI.

SergiiDmytruk (approved):

DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -540,6 +540,7 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciSerialParameters|$(PCI_SERIAL_PARAMETERS)
 
   gUefiCpuPkgTokenSpaceGuid.PcdCpuMaxLogicalProcessorNumber|$(MAX_LOGICAL_PROCESSORS)
+  gUefiCpuPkgTokenSpaceGuid.PcdFirstTimeWakeUpAPsBySipi|FALSE
```
mkopec, line 543: @miczyg1 shouldn't this go to `PcdsFixedAtBuild`? 
building qemu fails with:

```
build.py...
 : error 3000: Type [PatchableInModule] of PCD [gUefiCpuPkgTokenSpaceGuid.PcdFirstTimeWakeUpAPsBySipi] in DSC file doesn't match the type [FixedAtBuild] defined in DEC file.
    
 

- Failed -
Build end time: 16:09:22, Sep.26 2024
Build total time: 00:00:08

Failed!
make[1]: *** [Makefile:526: /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/FV/UEFIPAYLOAD.fd] Error 1
make: *** [payloads/external/Makefile.mk:171: build/UEFIPAYLOAD.fd] Error 2
```

SergiiDmytruk: I amended the commit to get https://github.com/Dasharo/edk2/commit/6b645e469f2e8d30be6823c1fb96b2a4427205cd.  Should have tested the build before approving.

PR #170, opened 2024-09-30
Runtime capsules

krystian-hebel:


DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
```diff
@@ -594,7 +600,7 @@ BlPeiEntryPoint (
   //
   LowMemorySize = PldMemInfo.UsableLowMemTop;
```
SergiiDmytruk, line 604: Any use of checking `LowMemorySize` for `0` here?  This doesn't seem likely, but debugging an issue caused by this won't be pleasant.

DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
```diff
@@ -342,7 +342,13 @@ MemInfoCallback (
         );
     } else if (Type == EFI_RESOURCE_SYSTEM_MEMORY) {
       if (Base < 0x100000000ULL) {
-        MemInfo->UsableLowMemTop = (UINT32)(Base + Size);
+        //
+        // Skip regions smaller than PeiMemSize as defined in BlPeiEntryPoint,
```
SergiiDmytruk:
```suggestion
        // Skip regions smaller than PeiMemSize as defined in BlPeiEntryPoint() below,
```
or define the size at the top.  I didn't get it's the entry point here at first because its name doesn't match the PEIM's name.

DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
```diff
@@ -604,7 +610,7 @@ BlPeiEntryPoint (
   //
   // Set cache on the physical memory
   //
-  MtrrSetMemoryAttribute (BASE_1MB, LowMemorySize - BASE_1MB, CacheWriteBack);
+  MtrrSetMemoryAttribute (BASE_1MB, PldMemInfo.SystemLowMemTop - BASE_1MB, CacheWriteBack);
```
SergiiDmytruk, line 617: Is commit message correct on this?  This is the top of the highest reserved memory region below 4G, not the highest region.  Also, the way it's obtained looks weird:
```c
SystemLowMemTop = ((UINT32)(Base + Size) + 0x0FFFFFFF) & 0xF0000000
```
```
(0xFED45000 + 0x0fffffff) & 0xf0000000 =  0x10ED44FFF & 0xf0000000 = 0
```

DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
```diff
@@ -431,7 +431,7 @@ DasharoEnableFUM (
   Status = gRT->SetVariable (
       DASHARO_VAR_FIRMWARE_UPDATE_MODE,
       &gDasharoSystemFeaturesGuid,
-      EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE,
+      EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_NON_VOLATILE,
```
SergiiDmytruk, line 434: Now `WarnIfFirmwareUpdateMode()` can't delete FUM variable because attributes don't match and it stays on forever.  Passing `0` for attributes in `SetVariable()` call causes variable deletion and avoids the need to synchronize attributes.

SergiiDmytruk (changes requested):

DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
krystian-hebel: Added an assert to check that it gives enough space instead of just testing against 0. The only possible invalid value is still 0 with the current code, but may be easier to understand by whoever reads it.

https://github.com/Dasharo/edk2/compare/b67eb08260c40f463eca88298e548cc70f8d3831..97938d04f1df2e564c1986fd34dc0dd00467cf54#diff-e11426c5786175fdf345eec361e1c716ab6792d8fddbd5b005e02ab589ad3d7bR605

DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
krystian-hebel: Defined it on top, in case it needs to be adjusted it will have to be done in only one place: https://github.com/Dasharo/edk2/compare/b67eb08260c40f463eca88298e548cc70f8d3831..97938d04f1df2e564c1986fd34dc0dd00467cf54#diff-e11426c5786175fdf345eec361e1c716ab6792d8fddbd5b005e02ab589ad3d7bR11

Also changed some literal values to `BASE_1MB` and `BASE_4GB` for readability.

DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
krystian-hebel: Yes, it didn't make sense, I don't really understand what that value was supposed to do. 256MB mask suggests it may be MMCONF, but it isn't fixed relatively to other MMIO. Reworked to actually check for MMIO.

https://github.com/Dasharo/edk2/compare/b67eb08260c40f463eca88298e548cc70f8d3831..97938d04f1df2e564c1986fd34dc0dd00467cf54#diff-e11426c5786175fdf345eec361e1c716ab6792d8fddbd5b005e02ab589ad3d7bL373-R383

DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
krystian-hebel: https://github.com/Dasharo/edk2/compare/b67eb08260c40f463eca88298e548cc70f8d3831..97938d04f1df2e564c1986fd34dc0dd00467cf54#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1137 and extended description in the commit message.

SergiiDmytruk (approved):

PR #172, opened 2024-10-04
to be squashed with "fix PEI memory range discovery code"

krystian-hebel:
In some cases, it isn't possible to differentiate between structures in RAM (e.g. coreboot tables) and memory reserved by System Agent (e.g. TSEG, GSM) based on memory region type. Both are reported as reserved memory and they may even be combined into one region, but their caching methods must be different. Because of that, MTRR setting code is no longer called at this point. This depends on coreboot setting memory caching attributes properly, which it already does.

### DIFF
--- a/DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
+++ b/DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
@@ -307,7 +307,7 @@ MemInfoCallback (
   IN VOID                         *Params
   )
 {
-  PAYLOAD_MEM_INFO        *MemInfo;
+  EFI_PHYSICAL_ADDRESS    *UsableLowMemTop;
   UINTN                   Attribute;
   EFI_PHYSICAL_ADDRESS    Base;
   EFI_RESOURCE_TYPE       Type;
@@ -322,11 +322,11 @@ MemInfoCallback (
              EFI_RESOURCE_ATTRIBUTE_WRITE_THROUGH_CACHEABLE |
              EFI_RESOURCE_ATTRIBUTE_WRITE_BACK_CACHEABLE;
 
-  MemInfo = (PAYLOAD_MEM_INFO *)Params;
-  Type    = MemoryMapEntry->Type;
-  Base    = MemoryMapEntry->Base;
-  Size    = MemoryMapEntry->Size;
-  Flag    = MemoryMapEntry->Flag;
+  UsableLowMemTop = (EFI_PHYSICAL_ADDRESS *)Params;
+  Type            = MemoryMapEntry->Type;
+  Base            = MemoryMapEntry->Base;
+  Size            = MemoryMapEntry->Size;
+  Flag            = MemoryMapEntry->Flag;
 
   if ((Base  < BASE_1MB) && ((Base + Size) > BASE_1MB)) {
     Size -= (BASE_1MB - Base);
@@ -348,7 +348,7 @@ MemInfoCallback (
         // region may be too small.
         //
         if (Size >= PEI_MEM_SIZE) {
-          MemInfo->UsableLowMemTop = (UINT32)(Base + Size);
+          *UsableLowMemTop = Base + Size;
         }
       } else {
         Attribute &= ~EFI_RESOURCE_ATTRIBUTE_TESTED;
@@ -373,14 +373,6 @@ MemInfoCallback (
         );
     } else if (Type == EFI_RESOURCE_MEMORY_MAPPED_IO) {
       BuildMemoryMappedIoRangeHob((EFI_PHYSICAL_ADDRESS)Base, Size);
-      //
-      // Find base of the lowest MMIO in range 1M - 4G.
-      //
-      if (Base < BASE_4GB) {
-        if (Base < MemInfo->SystemLowMemTop) {
-          MemInfo->SystemLowMemTop = Base;
-        }
-      }
     } else if (Type == EFI_RESOURCE_FIRMWARE_DEVICE) {
       BuildResourceDescriptorHob (
           EFI_RESOURCE_FIRMWARE_DEVICE,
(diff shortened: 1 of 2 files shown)


SergiiDmytruk:
Based on this log line from coreboot:

    top_of_ram = 0x47000000

And this one from EDK:

    5. 0000000047000000 - 00000000503FFFFF [05]

I think that coreboot changed memory map and newer definitions tend to just overwritten the old ones.  Not sure what these 148 MiB are for though.

krystian-hebel:
Partially accounted for:

```
[DEBUG]  SA MMIO resource: DSM      ->  base = 0x4c800000, size = 0x03c00000
[DEBUG]  SA MMIO resource: TSEG     ->  base = 0x4b800000, size = 0x00800000
[DEBUG]  SA MMIO resource: GSM      ->  base = 0x4c000000, size = 0x00800000
```

Not sure where `top_of_ram` comes from, this doesn't align with https://github.com/Dasharo/coreboot/blob/dasharo/src/mainboard/msi/ms7d25/romstage_fsp_params.c#L89 unless I'm missing something.

SergiiDmytruk:
`top_of_ram` gets printed in `fill_postcar_frame()` but that's before ramstage, so that part doesn't look relevant, but the same value could be used elsewhere after system agent is done and effectively drop those MMIO resources which fall into the range.

SergiiDmytruk (approved):
Given that coreboot replaces PEI, not touching MTRRs here makes sense.  Works on hardware as well.

miczyg1:
FYI On older EDK2 rbanch we had an option to disable MTRR programming: https://github.com/Dasharo/edk2/blob/dasharo/UefiPayloadPkg/UefiPayloadPkg.dsc#L228

> Not sure where top_of_ram comes from, this doesn't align with https://github.com/Dasharo/coreboot/blob/dasharo/src/mainboard/msi/ms7d25/romstage_fsp_params.c#L89 unless I'm missing something.

MMIOSize means to reserve 2.75G for MMIO, which means the initial boundary is 0x50000000, but that is for PCI. Belwo that are all the UMAs, stolen regions, etc. What coreboot gets is a bootloader's TOLUM, from which it further allocates TSEG and cbmem. So the top_of_ram printed by coreboot me already be below the cbmem.

DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
```diff
@@ -591,32 +582,22 @@ BlPeiEntryPoint (
...
   DEBUG ((DEBUG_INFO, "PeiMemBase: 0x%lx.\n", PeiMemBase));
   DEBUG ((DEBUG_INFO, "PeiMemSize: 0x%lx.\n", PEI_MEM_SIZE));
   Status = PeiServicesInstallPeiMemory (PeiMemBase, PEI_MEM_SIZE);
   ASSERT_EFI_ERROR (Status);
 
-  //
-  // Set cache on the physical memory
-  //
-  MtrrSetMemoryAttribute (BASE_1MB, PldMemInfo.SystemLowMemTop - BASE_1MB, CacheWriteBack);
-  MtrrSetMemoryAttribute (0, 0xA0000, CacheWriteBack);
```
miczyg1, line 618: Removal of this code could have been avoided by porting the change to disable MTRR programming:
https://github.com/Dasharo/edk2/blob/dasharo/UefiPayloadPkg/UefiPayloadPkg.dsc#L228

This feature was create for VP2410 initially, because changing MTRRs by EDK2 on this platform caused a hang. So if we ever make a new release for VP2410 using the rebased EDK2, it will likely not boot right now.

SergiiDmytruk: > This feature was create for VP2410 initially, because changing MTRRs by EDK2 on this platform caused a hang. So if we ever make a new release for VP2410 using the rebased EDK2, it will likely not boot right now.

The 2 sentences are seemingly in contradiction,  conditional use of `MtrrLibNull` with coreboot implies that EDK2 should set MTRRs sometimes.  What feature was added for VP2410, use of `MtrrLibNull` or `DISABLE_MTRR_PROGRAMMING` define?  EDIT: Or maybe the concern is that `MtrrSetMemoryAttribute()` is called in other places?

miczyg1: > The 2 sentences are seemingly in contradiction, conditional use of MtrrLibNull with coreboot implies that EDK2 should set MTRRs sometimes.

I just wanted the behavior to be modified on coreboot only at the time of developing it. Whether EDK2 should set MTRRs, I am not sure. It shouldn't need to change any caching attributes with UEFI Payload for coreboot. All resources are allocated and memory map is fixed already at coreboot stage. This may be different when a different bootloader is used though...

> What feature was added for VP2410, use of MtrrLibNull or DISABLE_MTRR_PROGRAMMING define? 

Both.

> EDIT: Or maybe the concern is that MtrrSetMemoryAttribute() is called in other places?

No concern here. Just saying that removing these MtrrSetMemoryAttribute calls would not be necessary if we had the `DISABLE_MTRR_PROGRAMMING` build option ported, in whatever form. If we didn't want the MTRRs to be set by EDK in the first place.

krystian-hebel: Are there cases where we would want EDK2 to set them?

miczyg1: > Are there cases where we would want EDK2 to set them?

I think there shouldn't be any cases with coreboot.

SergiiDmytruk: `EFI_CPU_ARCH_PROTOCOL` used in https://github.com/Dasharo/edk2/blob/dasharo/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c seems to set MTRRs too.  But if `MtrrLibNull`  was used without trouble before, sounds like it's OK to always use it in `DasharoPayloadPkg`.

I don't see much use of returning code which does nothing and can't even work properly though.  Its counterpart in `UefiPayloadPkg` doesn't exist anymore, so we might as well simplify this code to suit our needs and look less confusing.

miczyg1: For reference, this is how it is done in MrChromebox fork: https://github.com/MrChromebox/edk2/commit/8ea0ac749ff4cde85305544f79787b82f764b97e

SergiiDmytruk: Nice, I like that much more than the copy&paste variant.  Better switch to that one.

PR #175, opened 2024-11-04
.gitmodules: Update subhooks moduel URL to fix build

miczyg1:


mkopec (approved):

miczyg1:
Seems like there are mixed CRLF/LF endings in the file, thus more changes than one URL.

PR #192, opened 2024-11-26
OvmfPkg: Make OVMF buildable again

miczyg1:
May prove useful at some point.


OvmfPkg/Library/QemuParseLib/QemuParseLib.c
```diff
@@ -0,0 +1,119 @@
+/** @file
```
macpijan, line 1: Where does it come from? Is it restored from older commits? Pulled from upstream? Perhaps worth to note in the commit description.

macpijan (approved):
It builds just fine

OvmfPkg/Library/QemuParseLib/QemuParseLib.c
miczyg1: My bad, forgot. https://github.com/Dasharo/edk2/pull/192/commits/15f7619c240d52d1b63db921adb9f03a6b3b4969

PR #193, opened 2024-11-27
DasharoModulePkg: Add Fans Off feature

miczyg1:


miczyg1:
@mkopec I wonder if we want to hide the Fan off option on laptops. Rather not useful there.

mkopec:
@miczyg1 I think we want to hide that option on laptops

miczyg1:
> @miczyg1 I think we want to hide that option on laptops

Added in 0d915b890f8756980e6c967e82cb4e18b8ad5663

mkopec (approved):

PR #194, opened 2024-12-03
DasharoPayloadPkg: Switch to TPM PPI library from SecurityPkg

miczyg1:
The RAM backed TPM PPI based on coreboot RAM buffer and QEMU PPI library does not work reliably on all platforms. Switch to the TPM PPI implementation from SecurityPkg which uses UEFI variables to store TPM PPI buffer.

### DIFF
--- a/DasharoPayloadPkg/DasharoPayloadPkg.dsc
+++ b/DasharoPayloadPkg/DasharoPayloadPkg.dsc
@@ -326,8 +326,7 @@
 !if $(TPM_ENABLE) == TRUE
   Tpm12CommandLib|SecurityPkg/Library/Tpm12CommandLib/Tpm12CommandLib.inf
   Tpm2CommandLib|SecurityPkg/Library/Tpm2CommandLib/Tpm2CommandLib.inf
-  Tcg2PhysicalPresenceLib|OvmfPkg/Library/Tcg2PhysicalPresenceLibQemu/DxeTcg2PhysicalPresenceLib.inf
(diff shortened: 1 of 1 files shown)


miczyg1:
Tested on Protectli VP2410 and VP2420.

1. VP2410 previously didn't work (enable/disable SHA1 PCR bank). Now it works.
2. VP2420 previously worked (enable/disable SHA1 PCR bank) and still works.

Both firmwares were built from the same coreboot revision for the test after and before the change. Also used the same edk2 revision before the change.

The change causes the coreboot's TPM PPI buffer in ACPI not being used. However we have no tests to verify if TPM PPI works in OS currently... Switching to the PPI from SecurityPkg allows to get the TPM2 menu features working properly in firmware.

mkopec (approved):
We have confirmation from client that this solves their problem. BTW, do you know if this change also impacts TPM1 PPI? I recall testing it on m900_tiny and it didn't work at all

miczyg1:
> We have confirmation from client that this solves their problem. BTW, do you know if this change also impacts TPM1 PPI? I recall testing it on m900_tiny and it didn't work at all

@mkopec I think because TPM1.2 uses TcgPhysicalPresenceLib. The changes here modify the Tcg2PhysicalPresenceLib only. From what I see the UEFI Paylaod did not able TPM1.2 PPI at all. We would have to include SecurityPkg/Library/DxeTcgPhysicalPresenceLib INFs...

PR #195, opened 2024-12-04
Dynamic bmm

miczyg1:


miczyg1:
Did a quick test on OVMF and nothing exploded. The timeout also works as expected (F9 can reset it fro many level of setup browser without a need to enter BMM).

PR #198, opened 2024-12-12
Bring back TPM PPI in RAM

miczyg1:
Added more flushes and a stall before reset to get the flush or wbinvd hit the RAM.

Now with the VP66xx power cycles fixed: https://github.com/Dasharo/dasharo-blobs/pull/31 RAM backed for PPI works on VP6670 (the power cycle fix was not enough, the flushes in this PR are required as well). Similar fixes landed for VP32xx and VP2430 and the TPM PPI in RAM works there too.

Tested also on VP2410, where RAM backed did not want to work too. Unfortunately still doesn't work, even when request is sent from OS. The memory content of TPM PPI area looks random.


miczyg1:
@mkopec ping

mkopec (approved):

PR #199, opened 2024-12-18
[WIP] Add more feature flags

mkopec:
With everything set to FALSE, the payload is 140KB smaller.


miczyg1:
For some platforms it may be useful, e.g. Protectli, where we know that there is no SATA or EMMC or SD.

But the optimization is mainly on the binary size, not exactly boot time.

mkopec:
Sure, I just thought it might be valuable security-wise to remove as much unused code as possible

SCSI, (P)ATA and mouse related drivers seem like a waste of space for most of our platforms

PR #202, opened 2025-01-17
DasharoSystemFeaturesUiLib: Use callbacks to update CPU Throttling Threshold

miczyg1:
For some reason the CPU Throttling Threshold dsiplayed negative value when user changed CPU throttling, exited Power Management Options menu, went back to front page and then entered Power Management Options menu again.

Always calculate the threshold based on PCD and current throttling offset value. Add interactive flag to the options and patch the value in the callback.

### DIFF
--- a/DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c
+++ b/DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c
@@ -192,7 +192,6 @@ DasharoSystemFeaturesUiLibConstructor (
   PRIVATE_DATA(ShowPs2Option) = FixedPcdGetBool (PcdShowPs2Option);
   PRIVATE_DATA(Have2ndUart) = FixedPcdGetBool (PcdHave2ndUart);
   PRIVATE_DATA(ShowCpuThrottlingThreshold) = FixedPcdGetBool (PcdShowCpuThrottlingThreshold);
-  PRIVATE_DATA(CpuMaxTemperature) = FixedPcdGet8 (PcdCpuMaxTemperature);
   PRIVATE_DATA(ShowCpuCoreDisable) = FixedPcdGetBool(PcdShowCpuCoreDisable);
   PRIVATE_DATA(ShowCpuHyperThreading) = FixedPcdGetBool(PcdShowCpuHyperThreading);
   PRIVATE_DATA(WatchdogAvailable) = FixedPcdGetBool (PcdShowOcWdtOptions);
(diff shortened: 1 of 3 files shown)


krystian-hebel:
> Check if UEFI revocation list is up-to-date / check (pull_request) Failing after 7s 

Is it because of CVE-2024-7344?

mkopec:
@krystian-hebel [looks like it's because of CVE-2023-24932](https://uefi.org/revocationlistfile)

DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c
```diff
@@ -619,10 +618,29 @@ DasharoSystemFeaturesCallback (
 {
   EFI_STATUS                                 Status;
   EFI_INPUT_KEY                              Key;
+  DASHARO_SYSTEM_FEATURES_PRIVATE_DATA       *Private;
 
   Status = EFI_SUCCESS;
+  Private = DASHARO_SYSTEM_FEATURES_PRIVATE_DATA_FROM_THIS (This);
```
SergiiDmytruk, line 624: CI complains that this isn't used.  `PRIVATE_DATA()` uses `mDasharoSystemFeaturesPrivate` directly.

SergiiDmytruk:
> For some reason the CPU Throttling Threshold dsiplayed negative value when user changed CPU throttling, exited Power Management Options menu, went back to front page and then entered Power Management Options menu again.

This is worrisome, like there is a memory issue that will now affect something else since a field has been removed (e.g., `HybridCpuArchitecture` which is the next one).  Or is this an issue with how forms are implemented (now necessarily in our code)? Were you able to reproduce it on QEMU?  It didn't work for me on v0.2 with all the menus.

miczyg1:
> > For some reason the CPU Throttling Threshold dsiplayed negative value when user changed CPU throttling, exited Power Management Options menu, went back to front page and then entered Power Management Options menu again.
> 
> This is worrisome, like there is a memory issue that will now affect something else since a field has been removed (e.g., `HybridCpuArchitecture` which is the next one). Or is this an issue with how forms are implemented (now necessarily in our code)? Were you able to reproduce it on QEMU? It didn't work for me on v0.2 with all the menus.

I think it has to do with how `read` directive in VFR/IFR works and when it is invoked, rather than memory issue.

DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c
miczyg1: Right, I should change the access to CpuThrottlingOffset to use `Private` pointer.

miczyg1: Fixed: https://github.com/Dasharo/edk2/compare/e149e1ee6e13b71ed16a30aa27261d4c69eca3fe..27e4a00a26c5780ea53f8274f55c6dedb5496657

SergiiDmytruk (approved):
OVMF seems to work the same with and without these changes (variables are updated correctly), so seems like they at least don't break anything.

PR #203, opened 2025-01-28
DasharoPayloadPkg/SecureBootDefaultKeys: Update DBX to 14th Jan 2025 version

miczyg1:


miczyg1:
Ugh... The URL to DBX has changed to `https://uefi.org/sites/default/files/resources/dbxupdate_x64.bin`.  Maybe it would be wiser to obtain the DBX from https://github.com/microsoft/secureboot_objects/blob/main/PostSignedObjects/DBX/amd64/DBXUpdate.bin? WDYT @mkopec

mkopec:
Sounds good. I also thought we should set up a bot to create the PR automatically, like dependabot, haven't looked into it too much yet

mkopec (approved):

DasharoPayloadPkg/SecureBootDefaultKeys/DBXUpdate.bin
mkopec, line 1: Curious that the file is smaller, isn't dbx append only? @miczyg1

miczyg1: Hmm.. If the M$ PCA 2011 got revoked, maybe some signatures are no longer needed there?

mkopec: I guess so. Same difference in fwupd.

PR #204, opened 2025-01-28
Quiet and fast boot

miczyg1:
Works like a charm with Windows. The boot time in firmware is reduced from 11.1 seconds (effectively 8.1 seconds if 3second boot timeout is disregarded) to 4.9 seconds. The results may be checked in the Task Manager (Startup tab). Tested on MSI Z690-A DDR4 with RPL CPU.

The same time is achieved with Linux, with a small caveat. If the bootloader is GRUB, it will always try to read keystrokes, which means the console input will be connected by the firmware. However, the time consumed for connecting consoles is shifted to the bootloader itself. Summing it up, for Ubuntu, the boot time is as follows:

- 11.1 seconds firmware + ~7 seconds loader (fast boot disabled)
- 4.9 seconds firmware + ~10 seconds loader (fast boot enabled)

There is probably still more things can can be optimized. E.g. if the user ensures that each OS has ESP on the same disk as rootfs, we can skip connecting all storage medias unconditionally. In a case where ubuntu was installed alongside Windows, but on a different disk sharing the ESP, GRUB will fail to find the grub.cfg on the rootfs of the Linux distro.

~~Another small issue is that the fast boot only works with boot options created by the OS/bootloader. Boot options autodetected and autocreated by the firmware do not work (there seem to be a race conditions between refreshing the boot options and enumerating all devices after rivers are connected to controllers).~~

Ideas for future improvements:
1. Leverage `MdeModulePkg/Library/BootDiscoveryPolicyUiLib` and `MdeModulePkg/Universal/BootManagerPolicyDxe` to allow some customization of fast boot flow and the boot manager flow in general.


miczyg1:
Still I need to check how big is the impact of connecting storage controllers only  vs `EfiBootManagerConnectAll`.

Also need to test on headless system and a laptop with PS/2

pietrushnic:
@miczyg1 looks neat. Congratulations!

miczyg1:
> Still I need to check how big is the impact of connecting storage controllers only vs `EfiBootManagerConnectAll`.

Nah, `EfiBootManagerConnectAll` makes it over 9s. The autodetected entries are visible with that, but it is not worth it. Maybe we can add some delay to let the drives be enumerated properly. Or maybe I should recursively connect the drivers to drives....

miczyg1:
> > Still I need to check how big is the impact of connecting storage controllers only vs `EfiBootManagerConnectAll`.
> 
> Nah, `EfiBootManagerConnectAll` makes it over 9s. The autodetected entries are visible with that, but it is not worth it. Maybe we can add some delay to let the drives be enumerated properly. Or maybe I should recursively connect the drivers to drives....

Recursive connect helped

miczyg1:
Tested on VP6670 and no problems.

miczyg1:
Also tested on NS5x ADL. In Windows the Tasks Manager shows improvement from 7.2 seconds without fast boot to 3.9 seconds with fast boot.

The internal PS/2 keyboard also works in FW when firmware setup is requested from OS, so PS/2 should be fine.

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManager.vfr
```diff
@@ -51,6 +51,20 @@ formset
             step    = 0,
     endnumeric;
 
+    checkbox varid  = BmmData.QuietBoot,
+              questionid = FORM_QUIET_BOOT_ID,
+              prompt = STRING_TOKEN(STR_QUIET_BOOT_PROMPT),
+              help   = STRING_TOKEN(STR_QUIET_BOOT_HELP),
+              flags  = RESET_REQUIRED,
+    endcheckbox;
```
mkopec, line 59: Can we make it so that the Quiet Boot checkbox is checked and grayed out if Fast Boot is enabled? Since Fast Boot is essentially a superset of Quiet Boot

mkopec:
Seems to be working well on NovaCustom V540TU as well :+1:

mkopec:
Found some issues in further testing. On V540TU with windows 11, with fastboot enabled, the builtin keyboard stops working. I'm not sure why I didn't catch that before, maybe because the keyboard stayed initialized after reboot, but doesn't work after cold boot. Even when booting into the advanced startup menu or into the UEFI setup menu, the builtin keyboard does not work. Needed to use a USB keyboard to exit this state.

Windows boot manager also throws an error that it could not load winload.efi, when attempting to enter the advanced startup menu.

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManager.vfr
miczyg1: I don't see a reason to do so. The same way we would have to gray out timeout. And if we go further, we would have to select quiet boot and set timeout to 0  in the menu when fast boot is enabled. Then you would have to remember the setting from before enabling fast boot and restore it upon disabling. The overcomplication is the main reason I didn't go that way.

mkopec: Fair enough

miczyg1:
> Found some issues in further testing. On V540TU with windows 11, with fastboot enabled, the builtin keyboard stops working. I'm not sure why I didn't catch that before, maybe because the keyboard stayed initialized after reboot, but doesn't work after cold boot. Even when booting into the advanced startup menu or into the UEFI setup menu, the builtin keyboard does not work. Needed to use a USB keyboard to exit this state.
> 

I suspected there might be problems with PS/2. Similar things happened on MSI too (even before fast boot was implemented), that depending o nwarmboto or cold boot the PS/2 would or would not work in OS. Some initialization of PS/2 controller (not keyboard) seem to be required on boot, regardless of keyboard presence. 


> Windows boot manager also throws an error that it could not load winload.efi, when attempting to enter the advanced startup menu.

That's weird and have no idea where it may come from. I have not encountered that.

miczyg1:
> Found some issues in further testing. On V540TU with windows 11, with fastboot enabled, the builtin keyboard stops working. I'm not sure why I didn't catch that before, maybe because the keyboard stayed initialized after reboot, but doesn't work after cold boot. Even when booting into the advanced startup menu or into the UEFI setup menu, the builtin keyboard does not work. Needed to use a USB keyboard to exit this state.

For now forcing the extended verification for PS/2 init does not help. There must be some command which is missing. Still trying to lcoate it.

At current state (https://github.com/Dasharo/edk2/pull/204/commits/435efc9d8a9fbfe28a440d136fcd1f007e97fddf) the boot time for V540TND is 7.9s without fastboot and 4.5s with fast boot. Let's see how much loss there will be if I do full PS/2 init.

Also the V540TND manifests a weird issue. When I shutdown the laptop from Windows and then power it on again, I have 100% guaranteed BSOD with VIDEO_MEMORY_MANEGEMENT_INTERNAL code. So it makes it harder to test the original faulty scenario. However, PS/2 keyboard still does not work after the laptops resets after BSOD, so I think it may serve as a reproduction method still... @mkopec is this BSOD a known issue?

mkopec:
>  @mkopec is this BSOD a known issue?

It's not, please create an issue for this

miczyg1:
> > @mkopec is this BSOD a known issue?
> 
> It's not, please create an issue for this

https://github.com/Dasharo/dasharo-issues/issues/1239

miczyg1:
The PS/2 kBD issue does not reproduce on Linux. I suspect it is likely due to the boot flow. GRUB reads keystrokes which invoke the console connection process. Console connection process results in invoking the `InitKeyboard` indirectly via `ConsoleIn->ConInEx.Reset` inside `KbdControllerDriverStart`. This does not happen on Windows Boot Manager path leaving the controller uninitialized. So I will have to move some parts of the init to the `KbdControllerDriverSupported` I think, or to the `InitializePs2Keyboard` entry point (of course checking if the ports 0x60 and 0x64 reply with valid data, like it is done now).

miczyg1:
@mkopec pushed a fix for the non-working keyboard after shutdown. Can you please recheck?

Haven't checked the MSI with a pluggable PS/2 keyboard yet. Also there was an issue with detection I have added in the PlatformBootManager.

miczyg1:
> 7.9s without fastboot and 4.5s with fast boot

The loss is about 100ms for the PS/2 controller init added in commit https://github.com/Dasharo/edk2/pull/204/commits/83d5000c09a236c1319f51683a62e929b434fb21

Now I get 4.6 seconds in Task Manager.

miczyg1:
Found out the problem with PS/2 keyboard detection on my MSI Z690-A. The presence detection was done too early. Refactored it so that it is checked after consoles are connected on normal boot path (when the i8042 controller is already initialized). Also lowered the detection timeouts to half a second, as one second seemed a bit excessive.

Previously the PS/2 keyboard was not detected in time after pressing reset button, now it works perfectly. The boot performance of normal boot was 10.8 seconds (8 seconds if spamming enter whole time). When the keyboard is unplugged the boot is slightly longer, around 12.5 seconds (0.5s detection in keyboard driver + 0.25 seconds detection in Platform Boot Manager + a total of 1s of other timeouts happening in keyboard driver most likely). Initially I had 11.1 seconds of boot time on MSI (but it was in the lab unit, so different HW matrix).

miczyg1:
> @mkopec pushed a fix for the non-working keyboard after shutdown. Can you please recheck?

Ugh, I think it still doesn't work reliably. After a longer period of staying in S5, the keyboard just becomes unresponsive again in Windows. Moreover, I have discovered that Windows does not support PS/2 hotplug:  https://answers.microsoft.com/en-us/windows/forum/all/how-to-prevent-windows-disabling-ps2-port/579f4a65-02ab-482c-8eb6-8eec7fcb21fa (verified on my MSI Z690-A by hotplugging PS/2 keyboard after logging to the system, it does not appear in device manager). So I guess we might be doomed with skipping the PS/2 initialization on fastboot path. Best we can do is to leverage PcdSkipPs2Detect to run full init, assumingboard is connected on fastboot path.

Assuming that no support for PS/2 hotplug in WIndows is a holy truth, then on MSI when the keyboard is not present at boot there is no PS/2 keyboard in device manager. Otherwise, when it is connected at boot, it appears, regardless of fast boot option state or last power state (restart and power on both work ok).

However, on the NovaCustom V540TND the PS/2 keyboard is always present in Device Manager and it says the device is working properly... I am starting to believe that there is something different happening on EC side of PS/2 after simple reboot compared to the shutdown flow... @mkopec 

Some small details on what I saw when debugging the PS/2 controller state after reboot and power off:

1. After Reboot:

```
Keyboard data: 00
Keyboard status: 10
Keyboard command byte: 43
```

2. After power off and power on:

```
Keyboard data: 00
Keyboard status: 10
Keyboard command byte: 63
```

These are the dupms of the Data Port (0x60), Status Register (0x64)  and Configuration Byte (command 0x20).
The first two are read before any initialization happens. The Configuration Byte is dumped right before the Ps2KeyboardDxe writes 0x67 to it. So it is not always the same clean state. The EC keeps the translation bit set after shutdown but not after the reset...

Also one important thing, when I thought it is already working I had the AC plugged to the laptop. AC could cause the EC to hold some data after shutdown, which was not preserved when AC was not connected. Thus I got a false impression that I have fixed it. When I plug the AC it works, when unplug, it stops working... I'm starting to wonder if it was really necessary to commit this:  https://github.com/Dasharo/edk2/pull/204/commits/83d5000c09a236c1319f51683a62e929b434fb21

miczyg1:
TODO:

- look at this closer: https://review.coreboot.org/c/coreboot/+/46724
- check the translation on MSI after reboot and power on

pietrushnic:
> Found out the problem with PS/2 keyboard detection on my MSI Z690-A. The presence detection was done too early. Refactored it so that it is checked after consoles are connected on normal boot path (when the i8042 controller is already initialized). Also lowered the detection timeouts to half a second, as one second seemed a bit excessive.

Good news. Shouldn't it be linked to some issues? e.g. https://github.com/Dasharo/dasharo-issues/issues/418

EDIT: nvm, I read other messages, and it seems there are still issues.

miczyg1:
> > Found out the problem with PS/2 keyboard detection on my MSI Z690-A. The presence detection was done too early. Refactored it so that it is checked after consoles are connected on normal boot path (when the i8042 controller is already initialized). Also lowered the detection timeouts to half a second, as one second seemed a bit excessive.
> 
> Good news. Shouldn't it be linked to some issues? e.g. [Dasharo/dasharo-issues#418](https://github.com/Dasharo/dasharo-issues/issues/418)

Yes, the https://github.com/Dasharo/dasharo-issues/issues/418 will be fixed by commit https://github.com/Dasharo/edk2/pull/204/commits/b35b28df44392c253bc96f3ae18d22c362956385

And probably this issue will go away as well: https://github.com/Dasharo/dasharo-issues/issues/311

> 
> EDIT: nvm, I read other messages, and it seems there are still issues.

Yes, but that's only laptop-specific. Looks rather like an ill implementation on EC side.

miczyg1:
Repeated the register reads on V540TND with AC and without AC getting different results than before:

1. Reboot with AC:

```
Keyboard data: 00
Keyboard status: 10
Keyboard command byte: 73
```

2. Power off -> on with AC:

```
Keyboard data: 00
Keyboard status: 10
Keyboard command byte: 73
```

3. Reboot without AC:

```
Keyboard data: 00
Keyboard status: 10
Keyboard command byte: 70
```

4. Power off -> on without AC:

```
Keyboard data: 00
Keyboard status: 10
Keyboard command byte: 73
```

The only difference is the state of interrupts.

miczyg1:
Register reads on MSI:

1. Reboot:

```
Keyboard data: 00
Keyboard status: 00
Keyboard command byte: 30
```

2. Power off -> on:

```
Keyboard data: 00
Keyboard status: 00
Keyboard command byte: 30
```

3. Start from G3:

```
Keyboard data: 00
Keyboard status: 00
Keyboard command byte: 30
```

They are consistent. So the next step would be to make a clean state in PS2 on the EC side and see if the keyboard still works after reverting https://github.com/Dasharo/edk2/commit/83d5000c09a236c1319f51683a62e929b434fb21

miczyg1:
@mkopec I think I've found the culprit. After analyzing debug logs from EC for each possible boot flow with/without fastboot it looks like the faulty path lacks issuing KEYBOARD_KBEN (0xf4) command . On the EC side it enables the keyboard scanning. But it doesn't seem to be required on MSI desktops. 

It appears Windows PS/2 driver doesn't issue ht keyboard enable command too, thus it is not working... The PS/2 init flow from reset vector to Windows:

- normal boot path: https://paste.dasharo.com/?c78ddaaf806662f5#GeNwzLkYdgfTDKZmdSmVegUkHKdF9zXMZKFXj7dVBLPw
- fast bot path: https://paste.dasharo.com/?91e9c07c3f2f8da7#C2FEbNa1JcE2hnfZYaMH7YaHtgHavDhz5dXn526AxjAL

One can clearly see that on fast boot path there is no

```
kbc data: F4
  keyboard command
    enable scanning
```

Which means that boolean `kbscan_enabled` in EC code never gets set, so it purely depends on the EC keeping the state of the boolean. Of course the boolean was not kept when EC performed S5->G3 path (AC not connected on shutdown). With Linxu it always worked because GRUB issues ReadKeyStroke, causing the UEFI to perform full PS/2 keyboard init...

This commit fixes it: b26b83aa6c

Now the question is if we still need the `ConsoleIn->ConInEx.Reset` in `KbdControllerInit` from commit: `83d5000c09a236c1319f51683a62e929b434fb21` Will have to verify on both MSI and laptop, then it will be ready to go I think.

miczyg1:
This commit https://github.com/Dasharo/edk2/pull/204/commits/09f0b6a8d0e204908a62fad3b0dc3f69946a480b works as well on V560TNX, so if I can confirm it also works/does not break MSI boards, I will squash the commits adding the KBC init on fastboot path, otherwise, I will revert/drop it.

mkopec:
> Which means that boolean kbscan_enabled in EC code never gets set, so it purely depends on the EC keeping the state of the boolean.

Nice find. Do you think this (preserving kbscan_enabled across resets) should be considered a bug in the EC? Or maybe it should just default to `true`

miczyg1:
> > Which means that boolean kbscan_enabled in EC code never gets set, so it purely depends on the EC keeping the state of the boolean.
> 
> Nice find. Do you think this (preserving kbscan_enabled across resets) should be considered a bug in the EC?

Hmm. For sure this issue would be found faster by me if any PS/2 reset on EC side would actually restore default values of not only PS/2 registers but also any state variables inside the EC KBC implementation. If it was done properly, then I would see that fast boot breaks PS/2 keyboard on all paths (warm and cold, with and without AC). In this case either `kbscan_init` or `ps2_reset` for the first port should clear (set to false) `kbscan_enabled`. The same should happen with `kbc_first`, `kbc_second`, `kbc_translate` and `kbc_leds`. Also the `state_data` which holds the "Configuration Byte". It can be observed that some bits are preserved and some not, right here: https://github.com/Dasharo/edk2/pull/204#issuecomment-2665159642
The default at reset value of "Configuration Byte" should be `0x30` (judging by results on MSI desktop) meaning that both PS/2 ports are disabled.

But that's if we want to be super proper about it.

> Or maybe it should just default to true

Well, maybe it could? Still, we have the `kbc_first` which would block scanning too, if not set. And it seems to be default false. So it could work.

miczyg1:
Okay, so https://github.com/Dasharo/edk2/pull/204/commits/09f0b6a8d0e204908a62fad3b0dc3f69946a480b is optimizng too much for MSI desktops. WIndows refuses to init PS/2 keybaord if we issue KBEN command only on boot. So dropping this commit and leaving at https://github.com/Dasharo/edk2/pull/204/commits/b26b83aa6ca5231d5115792e54531c201b0ae13f where both laptops and MSI desktops still work well.

miczyg1:
@mkopec ready for final review and testing on your side.

mkopec:
@miczyg1 I just took the latest dasharo/coreboot rev, copied the V540TU defconfig and changed the edk2 rev to c149537db48463d2734371b8be6193bc7015097d , but it doesn't build:

```
Maintenance.c: In function 'InitializeBmmConfig':
/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c:1592:19: error: 'DASHARO_VAR_QUIET_BOOT' undeclared (first use in this function); did you mean 'DASHARO_VAR_ME_MODE'?
 1592 |                   DASHARO_VAR_QUIET_BOOT,
      |                   ^~~~~~~~~~~~~~~~~~~~~~
      |                   DASHARO_VAR_ME_MODE
"x86_64-elf-gcc"  -x c -E -P -DVFRCOMPILE --include FileExplorerLibStrDefs.h -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/FileExplorerLib -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/DEBUG -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Test/UnitTest/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Test/Mock/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Library/MipiSysTLib/mipisyst/library/include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Include/X64 -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Test/Mock/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BrotliCustomDecompressLib/brotli/c/include /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/FileExplorerLib/FileExplorerVfr.vfr > /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/OUTPUT/FileExplorerVfr.i
/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c:1593:20: error: 'gDasharoSystemFeaturesGuid' undeclared (first use in this function)
 1593 |                   &gDasharoSystemFeaturesGuid,
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
"VfrCompile" -l -n --string-db /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/OUTPUT/FileExplorerLibStrDefs.hpk --output-directory /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/DEBUG/. /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/OUTPUT/FileExplorerVfr.i
/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c:1605:19: error: 'DASHARO_VAR_FAST_BOOT' undeclared (first use in this function); did you mean 'DASHARO_VAR_WATCHDOG'?
 1605 |                   DASHARO_VAR_FAST_BOOT,
      |                   ^~~~~~~~~~~~~~~~~~~~~
      |                   DASHARO_VAR_WATCHDOG
```

mkopec:
CI is also failing

miczyg1:
> @miczyg1 I just took the latest dasharo/coreboot rev, copied the V540TU defconfig and changed the edk2 rev to [c149537](https://github.com/Dasharo/edk2/commit/c149537db48463d2734371b8be6193bc7015097d) , but it doesn't build:
> 
> ```
> Maintenance.c: In function 'InitializeBmmConfig':
> /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c:1592:19: error: 'DASHARO_VAR_QUIET_BOOT' undeclared (first use in this function); did you mean 'DASHARO_VAR_ME_MODE'?
>  1592 |                   DASHARO_VAR_QUIET_BOOT,
>       |                   ^~~~~~~~~~~~~~~~~~~~~~
>       |                   DASHARO_VAR_ME_MODE
> "x86_64-elf-gcc"  -x c -E -P -DVFRCOMPILE --include FileExplorerLibStrDefs.h -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/FileExplorerLib -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/DEBUG -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Test/UnitTest/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Test/Mock/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Library/MipiSysTLib/mipisyst/library/include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Include/X64 -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Test/Mock/Include -I/home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BrotliCustomDecompressLib/brotli/c/include /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/FileExplorerLib/FileExplorerVfr.vfr > /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/OUTPUT/FileExplorerVfr.i
> /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c:1593:20: error: 'gDasharoSystemFeaturesGuid' undeclared (first use in this function)
>  1593 |                   &gDasharoSystemFeaturesGuid,
>       |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~
> "VfrCompile" -l -n --string-db /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/OUTPUT/FileExplorerLibStrDefs.hpk --output-directory /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/DEBUG/. /home/coreboot/coreboot/payloads/external/edk2/workspace/Build/DasharoPayloadPkgX64/RELEASE_COREBOOT/X64/MdeModulePkg/Library/FileExplorerLib/FileExplorerLib/OUTPUT/FileExplorerVfr.i
> /home/coreboot/coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c:1605:19: error: 'DASHARO_VAR_FAST_BOOT' undeclared (first use in this function); did you mean 'DASHARO_VAR_WATCHDOG'?
>  1605 |                   DASHARO_VAR_FAST_BOOT,
>       |                   ^~~~~~~~~~~~~~~~~~~~~
>       |                   DASHARO_VAR_WATCHDOG
> ```

That's the oldest commit in the PR, not the topmost one.

mkopec:
sorry, i haven't had my coffee yet :| the latest commit builds

mkopec:
Ok, confirming it's working correctly on V540TU :+1:

mkopec (approved):

SergiiDmytruk:
> Okay, so [09f0b6a](https://github.com/Dasharo/edk2/commit/09f0b6a8d0e204908a62fad3b0dc3f69946a480b) is optimizng too much for MSI desktops. WIndows refuses to init PS/2 keybaord if we issue KBEN command only on boot. So dropping this commit and leaving at [b26b83a](https://github.com/Dasharo/edk2/commit/b26b83aa6ca5231d5115792e54531c201b0ae13f) where both laptops and MSI desktops still work well.

@miczyg1, and yet `MdeModulePkg/Bus/Isa/Ps2KeyboardDxe: Reduce PS/2 KBD init to minimum` has been merged.  Was it an accident?  I don't have any input on whether it works or not, was simply evaluating squashing that commit during rebase.

PR #209, opened 2025-02-12
Add support for FtdiUsbSerialDxe console redirection

filipleple:


DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -861,6 +861,7 @@
   MdeModulePkg/Bus/Usb/UsbMouseDxe/UsbMouseDxe.inf
   MdeModulePkg/Bus/Usb/UsbMassStorageDxe/UsbMassStorageDxe.inf
   MdeModulePkg/Bus/Usb/UsbMouseAbsolutePointerDxe/UsbMouseAbsolutePointerDxe.inf
+  FtdiUsbSerialDxe/FtdiUsbSerialDxe.inf
```
miczyg1: https://github.com/Dasharo/coreboot/pull/615#discussion_r1967490565

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
```diff
@@ -77,8 +77,10 @@ TerminalConInReset (
     TerminalDevice->DevicePath
     );
 
+  DEBUG ((DEBUG_INFO, "TERMINAL: TerminalConInReset() called\n"));
```
miczyg1: We probably don't need those debugs anymore right?

MdeModulePkg/Universal/Console/TerminalDxe/Terminal.c
```diff
@@ -521,15 +521,15 @@ TerminalDriverBindingStart (
     return Status;
   }
 
-  if (!IsHotPlugDevice (ParentDevicePath)) {
+  // if (!IsHotPlugDevice (ParentDevicePath)) {
```
miczyg1: I would like to avoid changing common code as much as possible. We should think how to get that working another way.

Like I mentioned, it should be platform boot manager specific to add this FTDI device as console at boot. I know it allows the FTDI device to be added as console in the boot maintenance manager menu and test it. But in the end, we probably want to have serial output on the FTDI device automatically on boot. So it will have to be done in platform boot manager anyways.

The way I see it is to add a pCD that configures whether FTDI device is to be added at boot, then loop through all handles with SimpleTExtIn and SimpleTextOut, got their device paths. Check if their device path contains USB node. If contains USB node, then it means it probably is a candidate for uSB serial. Check next node it if it is serial node. If yes, then append proper terminal GUID and add it to ConIn and ConOut. To be on the safe side, the loops would have to be done twice:
- loop through SimpleTextIn and add to ConIn
- loop through SimpleTextOut and add to ConOut

miczyg1 (changes requested):

MdeModulePkg/Universal/Console/TerminalDxe/Terminal.c
filipleple: This has just turned out to be unnecessary here, I've dropped the commit

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
filipleple: Right, dropped the commit

DasharoPayloadPkg/DasharoPayloadPkg.dsc
filipleple: Switched to including the top-level Pkg [here](https://github.com/Dasharo/edk2/pull/209/commits/58358c7efb4b1d556be0400854f44da0fe88cc81#diff-4c9d32b9f7d7dfb40b295a05a40ad511e85a6d9e499e7901abc79d21f9b18833R864)

miczyg1:
Now I would add a DEFINE in the DSC that will conditionally include the FDTI driver. I don't think we want that driver in every build.

filipleple:
[Added](https://github.com/Dasharo/edk2/pull/209/commits/cf6a1f07ebddbcebd7fa607cef943ec63a971f82) DEFINEs

PR #220, opened 2025-03-03
.github/workflows/dbx.yml: Don't fail the job if DBX is up to date

miczyg1:


mkopec (approved):

.github/workflows/dbx.yml
```diff
@@ -24,19 +29,23 @@ jobs:
         path: secureboot_objects
 
     - name: Check if DBX is out-of-date
+      id: check_hash
```
mkopec: is this needed?

miczyg1: Will be removed.

miczyg1: Final commit: https://github.com/Dasharo/edk2/pull/220/commits/80120f4f0bdf3cf6498a72c68b1318c448ccaf79

mkopec (approved):

PR #230, opened 2025-03-22
DasharoPayloadPkg: Implement support for initial bootloader CFR records

mkopec:
CFR is a new mechanism where bootloaders can provide a list of options to be parsed as an option menu. This EDK2 implementation will produce a setup menu at runtime.


mkopec:
Tested against coreboot branch dasharo-24.12, builds, boots and displays the Platform Setup Menu

PR #236, opened 2025-04-07
Revert "UefiCpuPkg: Produce EFI memory attributes protocol"

mkopec:
This protocol, introduced in UEFI v2.10 is known to break booting some scenarios: https://edk2.groups.io/g/devel/topic/110601533#msg121220

This includes booting Fedora 41 (either installer from usb or from disk, already installed).

We do not claim to support UEFI 2.10 spec, only 2.8 at this point. So disable the protocol and consider re-enabling once all software is updated to 2.10 spec and works with MAP.


mkopec:
Also see: https://github.com/tianocore/edk2/pull/10667

miczyg1 (approved):

PR #167, opened 2024-09-19
Capsule improve ux

krystian-hebel:
Position and size of progress bar produced by DisplayUpdateProgressLibGraphics
depends on the size of boot logo. As the logo can be customized by users,
a dummy logo with determinable dimensions, depending on the size of current
GOP mode, is created to make progress bar appear always in the same place.

PR #238, opened 2025-04-12
DasharoPayloadPkg/CapsuleChargerCheckDxe: Add AC check driver for updates

mkopec:


mkopec:
Untitled

Seems to work as intended.

miczyg1:
It looks like you used some scalable font here. Can it be scaled according to the display resolution IIUC? @mkopec

mkopec:
@miczyg1 Yes, exactly, but that's just copypasted from https://github.com/Dasharo/edk2/pull/167 , so credit goes to @krystian-hebel . I tried to split it out to a lib but couldn't do it in reasonable time (the single header libary is not conducive to being shared by different modules)

miczyg1:
> @miczyg1 Yes, exactly, but that's just copypasted from #167 , so credit goes to @krystian-hebel . I tried to split it out to a lib but couldn't do it in reasonable time (the single header libary is not conducive to being shared by different modules)

Why? You could add the array as C file, and have a second C file with lib constructor that will do the basic init, i.e. the `ssfn_load`; and contain other APIs. The array could be visible as `extern`. It doesn't necessarily have to be a header file.

DasharoPayloadPkg/CapsuleChargerCheckDxe/SSFN/README
SergiiDmytruk, line 1: Not sure what kind of issues you faced when trying to add it as an EDK library, but there should be no need to add includes as a copy, just move the headers to `DasharoPayloadPkg/Include/SSFN`.

DasharoPayloadPkg/CapsuleChargerCheckDxe/CapsuleChargerCheckDxe.c
```diff
@@ -0,0 +1,230 @@
...
+  //
+  // Clear the background. We don't know what was drawn there, if anything.
+  //
+  ZeroMem ((VOID *)GraphicsOutput->Mode->FrameBufferBase,
+           GraphicsOutput->Mode->FrameBufferSize);
+
+  //
+  // Print the error before resetting the platform.
+  //
+  Status = RenderTextCenteredAt("AC adapter check failed. Canceling update.",
```
SergiiDmytruk: I would mention automatic reboot in 10 seconds in the message.

SergiiDmytruk (changes requested):
These TODOs should probably be updated to say that capsules are responsible for battery checks:
https://github.com/Dasharo/edk2/blob/deb8d950fa0f0fb4215b104057840ec87caa04de/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c#L701
https://github.com/Dasharo/edk2/blob/deb8d950fa0f0fb4215b104057840ec87caa04de/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c#L1627

DasharoPayloadPkg/CapsuleChargerCheckDxe/CapsuleChargerCheckDxe.c
miczyg1: +1

Also, I would avoid using the word "fail", it's a bit scary. Having an unplugged AC is not a failure. I would rather display something nicer, e.g.


```
AC adapter is unplugged. Please plug in the AC and retry firmware update.
Cancelling update and rebooting in 10 seconds.
```

Or something among those lines.

miczyg1: We could also display a small logo depicting a power plug :) E.g.

https://thumbs.dreamstime.com/b/battery-plug-line-icon-charge-accumulator-sign-vector-battery-plug-line-icon-charge-accumulator-sign-electric-power-symbol-quality-272686961.jpg

krystian-hebel: > We could also display a small logo depicting a power plug :)

It may be possible to slightly modify SVG and store it as a glyph. If memory serves well, it has to use absolute coordinates with limited canvas size and can't use arcs. I tried to do this with Dasharo logo, but it is slightly rotated and didn't look good after quantization to smaller canvas:

image

There is also another limitation that may require splitting the image to smaller chunks, related to your previous question:

> Can it be scaled according to the display resolution IIUC?

Yes, but there is some max value for font size, 192 px IIRC. It is caused by a fact that some values are stored as `uint8_t` internally. The rest (192-255) is reserved for ascenders that are printed outside of font size (e.g. `Ś` or `Ü`).

miczyg1: >> We could also display a small logo depicting a power plug :) E.g.

> It may be possible to slightly modify SVG and store it as a glyph. If memory serves well, it has to use absolute coordinates with limited canvas size and can't use arcs. 

Just throwing ideas. If it is too complicated then we can do it at a later time. For now I will be satisfied with the change in the displayed string.

mkopec: Message updated in https://github.com/Dasharo/edk2/pull/238/commits/2d85ae5ad9ead6899b47ced0d5f1277519065aa6

DasharoPayloadPkg/CapsuleChargerCheckDxe/SSFN/README
mkopec: Done in https://github.com/Dasharo/edk2/pull/238/commits/86bac5fdbd9fe66fe0391f305d1e88985b0c646e

mkopec:
> These TODOs should probably be updated to say that capsules are responsible for battery checks:

Done in https://github.com/Dasharo/edk2/pull/238/commits/2f28eafa80407460064d81a0a1b5b41f053cde4e

SergiiDmytruk (approved):
LGTM, but needs history cleanup.  Rebase is necessary anyway due to the first commit having issues with the signature.

Also, commit message could say something like:
```
In the current implementation of capsule update, the capsules themselves
are responsible for checks like battery check on laptops.  This is achieved by
including the new DXE into capsules.  The DXE's entry point performs the check,
notifies the user and reboots the platform if AC isn't plugged.
```

miczyg1:
> LGTM, but needs history cleanup. Rebase is necessary anyway due to the first commit having issues with the signature.
> 
> Also, commit message could say something like:
> 
> ```
> In the current implementation of capsule update, the capsules themselves
> are responsible for checks like battery check on laptops.  This is achieved by
> including the new DXE into capsules.  The DXE's entry point performs the check,
> notifies the user and reboots the platform if AC isn't plugged.
> ```

+1

PR #244, opened 2025-04-24
MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c: Add BSD based systems

SebastianCzapla:
Needed for OSFV to boot systems in standard way.

Note that by deafult all of these systems install under efi/freebsd/loader.efi, so we need to change freebsd to opnsense/pfsense based on what derivative of FreeBSD we are using.

 
image


mkopec (approved):

miczyg1 (changes requested):
> so we need to change freebsd to opnsense/pfsense based on what derivative of FreeBSD we are using.

That means you are modifying the OS. It is not the right way to do it. The ESP scanning feature only consumes what OS exposes by default, not our custom modifications to the OS. If all BSD OSes expose `\EFI\\freebsd\\loader.efi`, then only the

```
  { L"\\EFI\\freebsd\\loader.efi",           L"FreeBSD (on %s)" },
```

is valid, however it will show FreeBSD on all BSD systems...

People will get confused on their systems because their OPNSense and pfSense installations will show up as FreeBSD. We can't do that. This is abusing/misusing the ESP scanning feature...

For Linux distros using the directory worked, because each distro used their own directory. Obviously, here it won't work, so you have to look for another way...

SebastianCzapla:
Closing as this is not appropriate solution to this issue. (Tracked here: https://github.com/Dasharo/dasharo-issues/issues/1306)

PR #253, opened 2025-06-16
Sb db owner guid

miczyg1:
Fix the owner GUID for Microsoft certificates and add missing MS Option ROM UEFI CA.


macpijan:
@mkopec would you take a look at this?

mkopec (approved):

PR #257, opened 2025-06-24
Handle FUM for capsule updates even when otherwise unsupported

SergiiDmytruk:
Protectli doesn't have FUM, but capsules rely on it to disable firmware protections.  We basically want the functionality without the menus, so I simply added boot mode check next to PCD check.  Same is done for coreboot in https://github.com/Dasharo/coreboot/pull/703

*ref: prot-1829*

### DIFF
--- a/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -1600,8 +1600,8 @@ PlatformBootManagerAfterConsole (
 
   WarnIfBatteryLow ();
   WarnIfRecoveryBoot ();
-  FUMEnabled = PcdGetBool (PcdShowFum) && WarnIfFirmwareUpdateMode ();
-
+  FUMEnabled = (PcdGetBool (PcdShowFum) || GetBootModeHob () == BOOT_ON_FLASH_UPDATE)
+            && WarnIfFirmwareUpdateMode ();
 
   BootLogoEnableLogo ();
 
--- a/SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c
+++ b/SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c
@@ -18,6 +18,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #include "AuthServiceInternal.h"
 #include <DasharoOptions.h>
+#include <Library/HobLib.h>
 
 ///
 /// Global database array for scratch
@@ -248,8 +249,10 @@ AuthVariableLibInitialize (
     SecureBootMode = SECURE_BOOT_MODE_DISABLE;
   }
 
-  // Disable Secure Boot if FUM enabled in the project and currently active
-  if (PcdGetBool (PcdShowFum) && !EFI_ERROR (Status))
+  //
+  // Disable Secure Boot if FUM enabled is in effect.
+  //
+  if ((PcdGetBool (PcdShowFum) || GetBootModeHob () == BOOT_ON_FLASH_UPDATE) && !EFI_ERROR (Status))
     SecureBootMode = SECURE_BOOT_MODE_DISABLE;
 
   Status = AuthServiceInternalUpdateVariable (


SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c
```diff
@@ -248,8 +249,10 @@ AuthVariableLibInitialize (
     SecureBootMode = SECURE_BOOT_MODE_DISABLE;
   }
 
-  // Disable Secure Boot if FUM enabled in the project and currently active
-  if (PcdGetBool (PcdShowFum) && !EFI_ERROR (Status))
+  //
+  // Disable Secure Boot if FUM enabled is in effect.
+  //
+  if ((PcdGetBool (PcdShowFum) || GetBootModeHob () == BOOT_ON_FLASH_UPDATE) && !EFI_ERROR (Status))
```
krystian-hebel, line 255: Wouldn't the opposite be needed for https://github.com/Dasharo/dasharo-issues/issues/1075?

How does the firmware behave without this modification and SB enabled? I expect FMP drivers to be always denied (even when they would be properly signed) due to [GetImageType returning `IMAGE_UNKNOWN`](https://github.com/Dasharo/edk2/blob/10486fbe47856ca9ed6acbe2a7cba3c6ac8455dd/SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c#L1709), but maybe I'm missing something.

SergiiDmytruk: > Wouldn't the opposite be needed for https://github.com/Dasharo/dasharo-issues/issues/1075?

Possibly, unless something other than SB is used for authentication.

> How does the firmware behave without this modification and SB enabled?

Here's how (in QEMU):
```
Updating the firmware ......
ProcessThisCapsuleImage - 0x7ECF4000
ProcessCapsuleImage for FmpCapsule ...
ValidateFmpCapsule ...
ValidateFmpCapsule - Success
ProcessFmpCapsuleImage ...
FmpCapsule: LoadImage ...
[Security] 3rd party image[0] can be loaded after EndOfDxe: MemoryMapped(0x3,0x7ECF4040,0x7ED10D3F).
DxeImageVerificationLib: Image is not signed and SHA1 hash of image is not found in DB/DBX.
The image doesn't pass verification: MemoryMapped(0x3,0x7ECF4040,0x7ED10D3F)
FmpCapsule: LoadImage - Access Denied
Driver Return Status = Access Denied
ProcessFmpCapsuleImage - Access Denied
ProcessThisCapsuleImage - Access Denied
ERROR: C00000002:V03818005 I0 6D33944A-EC75-4855-A54D-809C75241F6C
Capsule process failed!
```
So, `Policy = DENY_EXECUTE_ON_SECURITY_VIOLATION;` and look up in DB/DBX.

krystian-hebel: OK, so this is another thing to keep in mind when the authentication is being implemented.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1600,8 +1600,8 @@ PlatformBootManagerAfterConsole (
 
   WarnIfBatteryLow ();
   WarnIfRecoveryBoot ();
-  FUMEnabled = PcdGetBool (PcdShowFum) && WarnIfFirmwareUpdateMode ();
-
+  FUMEnabled = (PcdGetBool (PcdShowFum) || GetBootModeHob () == BOOT_ON_FLASH_UPDATE)
```
krystian-hebel, line 1603: This would introduce a vulnerability.

`WarnIfFirmwareUpdateMode()` does two important things here:
- asks for confirmation, so that the user is aware of security implications of having protections disabled (this is silenced when `BOOT_ON_FLASH_UPDATE`, which is fine since this mode always ends in reboot before giving control to user)
- removes FUM UEFI variable, and creates a volatile one in its place, so after next reboot FUM is disabled

In old code, this function is invoked when FUM is shown in menu. In new code, it is called when FUM is shown in menu, or when capsules **are present in the memory**. Putting [changes in coreboot](https://github.com/Dasharo/coreboot/pull/703/files#diff-9219efc7069e199d2a4ae6d648d4f9986874aaed38374fa1f2567f9971d5a305R180) in contrast, existence of FUM UEFI var results in protections disabled when FUM is shown in menu, or when capsules **are enabled**.

This means that when capsules are enabled, but FUM isn't, it is possible to manually create nonvolatile FUM UEFI variable and all of the protections will be permanently disabled, with no indication to the user.

This has to be changed somehow. I can think of few solutions:
1. Make FUM a requirement of capsule updates
2. Move conditions from edk2 to coreboot - not sure if feasible, some decisions may have to be made before capsule in memory can be detected
3. Move conditions from coreboot to edk2, so `WarnIfFirmwareUpdateMode()` is always called if any of those two features are enabled. With no other changes, in the scenario described above, this would show the FUM popup and ask for confirmation. This would be acceptable from security point of view, but it would mean that FUM is practically available, just not exposed in the menu. It would be basically the same as 1. (with one write to efivarfs instead of navigatin through the menu), but for free.
4. Same as 3., but with `WarnIfFirmwareUpdateMode()` modified so it reboots when `PcdShowFum` is false and there are no capsules in the memory. While at it, there is a [memory leak](https://github.com/Dasharo/edk2/blob/10486fbe47856ca9ed6acbe2a7cba3c6ac8455dd/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c#L1184), the check for capsule should be made before this event is created.

There may be other solutions I haven't thought about.

krystian-hebel (changes requested):

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
SergiiDmytruk: > This means that when capsules are enabled, but FUM isn't, it is possible to manually create nonvolatile FUM UEFI variable and all of the protections will be permanently disabled, with no indication to the user.

Why permanently?  `WarnIfFirmwareUpdateMode()` still "removes FUM UEFI variable, and creates a volatile one in its place, so after next reboot FUM is disabled".  What's the difference compared to when FUM is supported?

I realized the asymmetry but didn't check for `CapsuleUpdateData` in coreboot because if one UEFI variable can be faked, the same applies to all of them and complete parsing of capsules might not work early enough when `options.c` is used.  Maybe I'm missing something, but it seemed that the proposed changes should have the same effects as FUM.

krystian-hebel: > Why permanently? `WarnIfFirmwareUpdateMode()` still "removes FUM UEFI variable, and creates a volatile one in its place, so after next reboot FUM is disabled". What's the difference compared to when FUM is supported?

It would, if `WarnIfFirmwareUpdateMode()` was always called. This doesn't happen when `PcdShowFum` is set to false and there are no capsules in the memory.

BeataZdunczyk: @wiktormowinski Let's enable FUM. The FUM feature should be a dependency for implementing the Capsule Update feature.

@krystian-hebel, @SergiiDmytruk cc

wiktormowinski: @BeataZdunczyk ok i built it from: https://github.com/Dasharo/coreboot/pull/703
but with edk2 revision from dasharo branch (e9a1d19bfa9fb5d6ef52d1dfd96e544c0f35f248) and not this one
and with FUM enabled
and I can confirm: it works!

No need to look for the workaround ig.
@krystian-hebel, @SergiiDmytruk cc

PR #258, opened 2025-06-25
DasharoPayloadPkg: Add VirtIO drivers build for QEMU

### LINKED ISSUE #901: VirtIO drivers for QEMU Q35 releases
**Bounty: $200**

### The problem you're addressing (if any)

UEFI payload in release v0.2.0 for Q35 doesn't include VirtIO drivers. Because of that, bootable disks have to be mounted using different interfaces, which are generally slower in virtualized environment.

### Describe the solution you'd like

(Conditionally?) add VirtIO drivers to DasharoPayloadPkg.

### Where is the value to a user, and who might that user be?

Slightly better performance of I/O operations.

### Describe alternatives you've considered

_No response_

### Additional context

* VirtIO driver enabled in edk2 build for QEMU
  - or possibly for all builds, but it should be loaded under QEMU only
* VirtIO test suite might go into [this directory](https://github.com/Dasharo/open-source-firmware-validation/tree/develop/dasharo-compatibility) and be guarded by flags such as [this one](https://github.com/Dasharo/open-source-firmware-validation/blob/develop/lib/network.robot#L17) to make sure it's executed on QEMU machines only

### Definition of done

* VirtIO edk2 driver is loaded when running on QEMU
* VirtIO driver is **not loaded** when running on some hardware platform
* Attached logs of newly developed test, proving VirtIO driver is visible in firmware

aronowski:
Add VirtIO drivers build for QEMU. Conditionally include module definitions for VirtIO drivers and their dependencies for q35 mainboard (`[Components.X64]` only).

# Description

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

As a proof that the drivers are built for QEMU and active, these coreboot runs are attached:
- https://github.com/aronowski/coreboot/actions/runs/15879961202
- https://github.com/aronowski/coreboot/actions/runs/15880467829

as well as [the suggestion by @philipanda](https://github.com/Dasharo/dasharo-issues/issues/901#issuecomment-2979606056):

> The DoD says to attach the `logs of newly developed test, proving VirtIO driver is visible in firmware`. The new test is not required to test the changes in EDK2 & coreboot, because the `esp-scanning.robot` test does that incidentally.

In the [first run](https://github.com/aronowski/coreboot/actions/runs/15879961202) you can see in the build_q35 jobs, that the relevant VirtIO drivers are built for [X64] and tested with [my OSFV fork](https://github.com/aronowski/open-source-firmware-validation/commit/11a313dfcf86ec1016be7f5981b8847b5972c5c3), where `if=virtio` is used instead of `if=ide`, and the `esp-scanning.robot` test suite passes.

In the [second run](https://github.com/aronowski/coreboot/actions/runs/15880467829) I intentionally broke the test_q35 job by trying to run `esp-scanning.robot` using my OSFV revision with forced `if=virtio` and [the edk2 revision prior to this PR (without VirtIO drivers)](https://github.com/aronowski/coreboot/commit/6bb7c73ae093abc751fe156b911bcad1d9365dbb), further proving that the drivers get built and are visible in firmware as per the current PR.

If, despite the suggestion and the justification above, the `esp-scanning.robot` test suite is not sufficient to fulfill the DoD, and a new test is indeed needed, I can provide one in the future.

## Integration Instructions

This is the first PR to partially resolve https://github.com/Dasharo/dasharo-issues/issues/901, [as suggested](https://github.com/Dasharo/dasharo-issues/issues/901#issuecomment-2979606056):

> Two PRs are needed and the one to EDK2 should be merged first.

[This](https://github.com/aronowski/edk2/actions/runs/15878283324) is the revision of edk2 with a modified GitHub Workflow, so that the pushes to my branch are built, as a proof that it builds fine in the first place.

Integration-wise, this PR shall be self-sufficient and non-breaking in the context of edk2, i.e. the coreboot and OSFV changes shall be integrated in the future after some minor changes, like commit titles.

### DIFF
--- a/DasharoPayloadPkg/DasharoPayloadPkg.dsc
+++ b/DasharoPayloadPkg/DasharoPayloadPkg.dsc
@@ -189,6 +189,11 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsic.inf
   PciCf8Lib|MdePkg/Library/BasePciCf8Lib/BasePciCf8Lib.inf
   PciLib|DasharoPayloadPkg/Library/BasePciLibPciExpress/BasePciLibPciExpress.inf
+!if $(QEMU_PLATFORM) == TRUE # Needed for Virtio10.inf in QEMU
+  PciCapLib|OvmfPkg/Library/BasePciCapLib/BasePciCapLib.inf
+  PciCapPciIoLib|OvmfPkg/Library/UefiPciCapPciIoLib/UefiPciCapPciIoLib.inf
+  VirtioLib|OvmfPkg/Library/VirtioLib/VirtioLib.inf
+!endif
   PciExpressLib|DasharoPayloadPkg/Library/BasePciExpressLib/BasePciExpressLib.inf
   PciSegmentLib|MdePkg/Library/BasePciSegmentLibPci/BasePciSegmentLibPci.inf
   PeCoffLib|MdePkg/Library/BasePeCoffLib/BasePeCoffLib.inf
@@ -394,6 +399,7 @@
   LaptopLidLib|DasharoPayloadPkg/Library/LaptopLidLib/LaptopLidLibNull.inf
 !endif
 !endif
+OrderedCollectionLib|MdePkg/Library/BaseOrderedCollectionRedBlackTreeLib/BaseOrderedCollectionRedBlackTreeLib.inf
 
 [LibraryClasses.IA32.SEC]
   DebugLib|MdePkg/Library/BaseDebugLibNull/BaseDebugLibNull.inf
@@ -735,13 +741,19 @@
   DebugLib|MdeModulePkg/Library/PeiDxeDebugLibReportStatusCode/PeiDxeDebugLibReportStatusCode.inf
 
 [Components.X64]
+!if $(QEMU_PLATFORM) == TRUE
+  OvmfPkg/VirtioPciDeviceDxe/VirtioPciDeviceDxe.inf
+  OvmfPkg/Virtio10Dxe/Virtio10.inf
+  OvmfPkg/VirtioBlkDxe/VirtioBlk.inf
+  OvmfPkg/VirtioScsiDxe/VirtioScsi.inf
+!endif
+
   #
   # DXE Core
   #
   MdeModulePkg/Core/Dxe/DxeMain.inf {
     <LibraryClasses>
       NULL|MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf
-      OrderedCollectionLib|MdePkg/Library/BaseOrderedCollectionRedBlackTreeLib/BaseOrderedCollectionRedBlackTreeLib.inf
   }
 
   #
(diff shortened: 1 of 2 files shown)


philipanda (commented):
Great that you've run and linked all of these CI runs here, it helps a lot.
I've built the firmware manually to make sure and the changes work fine.
The code looks good, my only suspicion is if all of the newly added drivers are required for virtio disks, like:

```
 OvmfPkg/VirtioRngDxe/VirtioRng.inf
 OvmfPkg/VirtioNetDxe/VirtioNet.inf 
```

I've checked that they are not required for virtio disks to work by building and running the OSFV esp-scanning test. I am not sure if we should add these as we are not checking if they work.

philipanda (changes requested):
I'd like to ask you to remove any drivers which are not required for VirtIO disks, like VirtioRNG and VirtioNET for now. Unless you'd like to suggest tests for their functionality. I think the author of [the Issue](https://github.com/Dasharo/dasharo-issues/issues/901) simply didn't take into account that there are might be more uses of VirtIO than disks

aronowski:
Removed the modules, which weren't needed for the test suite to pass. [Here's a run, where the test_q35 jobs pass](https://github.com/aronowski/coreboot/actions/runs/15908874764/job/44871337023).

As far as the recent workflow runs go, both VirtioPciDeviceDxe.inf and Virtio10.inf appear to be needed, otherwise the job fails:
- https://github.com/aronowski/coreboot/actions/runs/15908364647/job/44870001083
- https://github.com/aronowski/coreboot/actions/runs/15907555938/job/44867154855

philipanda (approved):
I think everything's good now. We'll wait for @macpijan with merging. You can assume it's merged and move https://github.com/Dasharo/coreboot/pull/704 out of draft after changing the references to EDK2

macpijan:
Would you mind taking a look @SergiiDmytruk ? Just to make sure nothing was missed in the edk2 changes.

DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -404,6 +411,11 @@
   ReportStatusCodeLib|MdePkg/Library/BaseReportStatusCodeLibNull/BaseReportStatusCodeLibNull.inf
   PerformanceLib|MdeModulePkg/Library/PeiPerformanceLib/PeiPerformanceLib.inf
 
+[LibraryClasses.common]
+!if $(QEMU_PLATFORM) == TRUE
+  VirtioLib|OvmfPkg/Library/VirtioLib/VirtioLib.inf
+!endif
```
SergiiDmytruk: This can be added to `[LibraryClasses]`.

DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -394,6 +398,9 @@
   LaptopLidLib|DasharoPayloadPkg/Library/LaptopLidLib/LaptopLidLibNull.inf
 !endif
 !endif
+!if $(QEMU_PLATFORM) == TRUE # Needed for Virtio10.inf in QEMU
+  OrderedCollectionLib|MdePkg/Library/BaseOrderedCollectionRedBlackTreeLib/BaseOrderedCollectionRedBlackTreeLib.inf
+!endif
```
SergiiDmytruk: This doesn't need to be conditional, the library is used in 2 other places in this file, so we can drop it from there and add here unconditionally.

SergiiDmytruk (commented):
There is also `OvmfPkg/VirtioFsDxe/VirtioFsDxe.inf`, but it's not really about disk images, so I guess we don't really need it.

Changes can also be optionally simplified.

DasharoPayloadPkg/DasharoPayloadPkg.dsc
aronowski: Moved to `[LibraryClasses]`.

DasharoPayloadPkg/DasharoPayloadPkg.dsc
aronowski: Removed those 2 other references and left this one unconditionally. [The test_q35 jobs pass](https://github.com/aronowski/coreboot/actions/runs/16051027185/job/45293983141).

SergiiDmytruk (approved):

PR #259, opened 2025-06-26
Sb wizard option

miczyg1:


miczyg1:
Operation confirmed with tests: https://github.com/Dasharo/open-source-firmware-validation/pull/915

PR #260, opened 2025-06-26
Fix file explorer empty statement

miczyg1:
Fixes the bug where the test have to press an extra arrow in the File Explorer:

```
Execute File In File Explorer
    [Documentation]    Execute the given file (with ENTER key)
    [Arguments]    ${target_file}
    # 1. Select desired file
    ${files}=    Get Submenu Construction
    Log    ${files}
    ${index}=    Get Index Of Matching Option In Menu    ${files}    ${target_file}
    # FIXME: We must add 1 due to empty selecatble space in File Manager
    Press Key N Times And Enter    ${index}+1    ${ARROW_DOWN}
```

There was an empty string text statement between the directory/file creation and the list of file/directories, which was selectable. Change this statement to subtitle, which is not selectable, to fix the problem.

Also disallow highlighting empty strings or strings containing a single space character. This will additionally allow to put empty statements into forms that are not supposed to highlight any entry by default when entered.

### DIFF
--- a/MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
+++ b/MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
@@ -2080,11 +2080,14 @@ DisplayMenuString (
     return;
   }
 
-  //
-  // Print the highlight menu string.
-  // First print the highlight string.
-  //
-  SetDisplayAttribute (MenuOption, TRUE);
+  // Only highlight non-empty strings
+  Highlight = StrLen(MenuOption->Description) != 0;
+  // Sometimes empty strings are defined as single character strings with space
+  if (StrLen(MenuOption->Description) == 1 && MenuOption->Description[0] == L' ') {
+    Highlight = FALSE;
+  }
+
+  SetDisplayAttribute (MenuOption, Highlight);
   Length = PrintStringAt (Col, Row, String);
 
   //
--- a/MdeModulePkg/Library/FileExplorerLib/FileExplorer.c
+++ b/MdeModulePkg/Library/FileExplorerLib/FileExplorer.c
@@ -1347,10 +1347,11 @@ LibUpdateFileExplorePage (
         EFI_IFR_FLAG_CALLBACK,
         (UINT16)(mNewFolderQuestionId++)
         );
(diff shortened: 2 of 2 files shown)


MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -2080,11 +2080,14 @@ DisplayMenuString (
...
-  //
-  SetDisplayAttribute (MenuOption, TRUE);
+  // Only highlight non-empty strings
+  Highlight = StrLen(MenuOption->Description) != 0;
+  // Sometimes empty strings are defined as single character strings with space
+  if (StrLen(MenuOption->Description) == 1 && MenuOption->Description[0] == L' ') {
+    Highlight = FALSE;
+  }
+
+  SetDisplayAttribute (MenuOption, Highlight);
```
SergiiDmytruk, line 2090: That extra line in file explorer was annoying and I'm glad it's gone, but this highlighting is what made it possible to see what was going on by knowing where the cursor is.  Are you sure this won't create places where the cursor becomes invisible?  Maybe that's why a single space was used instead of some empty strings.

miczyg1: As explained in the description it is intentional to have the cursor hidden:

```
Also disallow highlighting empty strings or strings containing a single space character. This will additionally allow to put empty statements into forms that are not supposed to highlight any entry by default when entered.
```

I need this for Sovereign Boot Wizard, where the customer requests no entry to be higlighted by default when the form is entered. Also the problem is that the form browser does not allow to put the cursor focus on an option if its string description is empty (just starts with `\0`). So this one space check is required.

SergiiDmytruk: I wasn't questioning the change or a use case for it.  My point is that this affects all forms and thus can hide the cursor where it better be visible.  However, after grepping for `" "` and checking things in QEMU this doesn't seem to be that common, so we can fix any such cases as we find them and if we find them.

SergiiDmytruk (approved):

PR #262, opened 2025-07-01
Ftdi autoselect

filipleple:
Solves the problem where you manually have to select the FTDI from the console options menu, and then it's subject to getting deselected if the exact device path doesn't persist across reboots.

Now we

* look for devices with both UsbIo and SimpleText protocols
* Check the VID and PID if that's a FTDI device (supported by the driver we use)
* Attach it to the console-related variables

in PlatformBootManager.c

### DIFF
--- a/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -18,6 +18,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/HobLib.h>
 #include <Coreboot.h>
 #include <DasharoOptions.h>
+#include <Protocol/UsbIo.h>
 
 EFI_GUID mBootMenuFile = {
   0xEEC25BDC, 0x67F2, 0x4D95, { 0xB1, 0xD5, 0xF8, 0x1B, 0x20, 0x39, 0xD1, 0x1D }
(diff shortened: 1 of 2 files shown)


mkopec:
do you still have to set `Serial Console Redirection` to enabled to enable the uart adapter?

filipleple:
> do you still have to set `Serial Console Redirection` to enabled to enable the uart adapter?

yup, all ftdi-related messages are missing from cbmem with serial redir disabled, and the device is non-functional in FW

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -27,6 +28,8 @@ EFI_GUID mBootMenuFile = {
 extern BOOLEAN mQuietBoot;
 extern BOOLEAN mFastBoot;
 
+STATIC VOID RegisterFtdiUsbUart(VOID);
```
krystian-hebel: Forward declaration can be avoided by moving the function definition here.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1858,3 +1864,113 @@ PlatformBootManagerUnableToBoot (
...
+      DevPathStr = ConvertDevicePathToText (FullDevicePath, FALSE, FALSE);
+      if (DevPathStr != NULL) {
+        DEBUG ((DEBUG_INFO, "Registering FTDI terminal path: %s\n", DevPathStr));
+        FreePool (DevPathStr);
+      }
+
+      EfiBootManagerUpdateConsoleVariable (ConIn, FullDevicePath, NULL);
+      EfiBootManagerUpdateConsoleVariable (ConOut, FullDevicePath, NULL);
+      EfiBootManagerUpdateConsoleVariable (ErrOut, FullDevicePath, NULL);
+
```
krystian-hebel: Have you tried calling `ConIn->Reset()` at this point? It may fix the problem with registering setup key on first boot.

krystian-hebel (changes requested):

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1619,6 +1739,11 @@ PlatformBootManagerAfterConsole (
 
     // With fast boot, we can't call ConnectAll as it would connect all consoles.
     EfiBootManagerConnectAll ();
+    
+    // Detect and register FTDI USB-UART converters
+    // The FTDI can only be detected after all protocols are bound thanks to
+    // the ConnectAll, but then the consoles have to be reconnected.
+    if (RegisterFtdiUsbUart()==EFI_SUCCESS) EfiBootManagerConnectAll ();
```
filipleple: @mkopec @krystian-hebel so this finally worked. Is a workaround like this acceptable, or maybe you know of just a line from `ConnectAll` that has to be called here?

filipleple: OK nvm, connecting only the FTDI path instead [here](https://github.com/Dasharo/edk2/compare/9151adcc0121caca0b1fbe47eae4fc3d5a0df429..7b8116b14098ffda2be2c6fa765955b0e1bb9ddc#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1750)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [done](https://github.com/Dasharo/edk2/pull/262/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR32)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: Hmm, this didn't end up being the solution

krystian-hebel (approved):

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
krystian-hebel: I've approved the rest, but I'd appreciate if Michał would take another look at this particular part.

mkopec (approved):

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1619,7 +1741,16 @@ PlatformBootManagerAfterConsole (
 
     // With fast boot, we can't call ConnectAll as it would connect all consoles.
     EfiBootManagerConnectAll ();
-  }
+    
+    // Detect and register FTDI USB-UART converters
+    // The FTDI can only be detected after all protocols are bound thanks to
+    // the ConnectAll, but then it has to be connected separately
+    EFI_DEVICE_PATH_PROTOCOL *FtdiPath;
+    if (EFI_SUCCESS == RegisterFtdiUsbUart(&FtdiPath)) {
```
krystian-hebel, line 1744: This is too late. Screens produced by all of the `WarnIf*` functions aren't visible on serial.

---

I wonder if `UpdatePs2KeyboardConIn` also isn't called too late. This could be tested by trying to use FUM on MSI with PS/2 keyboard.

krystian-hebel (dismissed):

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
miczyg1: it has to be done after EfiBootManagerConnectAll because before that, the drivers are not connected -> USB devices not enumerated

I think the same happens with PS/2 keyboard. Only after EfiBootManagerConnectAll is called, the driver is started and initializes the keyboard. UpdatePs2KeyboardConIn is supposed to remove the keyboard from ConIn if it is not connected to the system, because the way how PS2 keyboard driver works the PlatformBootManagers must add the PS2 keyboard to ConIn.

krystian-hebel: Then lets look at it from different point: would something break if `WarnIf*` were instead moved later, after all of what you mentioned happens?

filipleple: @krystian-hebel https://github.com/Dasharo/edk2/pull/262/commits/895389bb1fc8191fa66bf7ff4c934216947d5ecc I've tried moving the warnifs past the ConnectAll. Looks like nothing broke and now the FUM popup shows up over FTDI.

PR #263, opened 2025-07-03
Setup menu color customization

philipanda:
# Description

Extracting hardcoded colors from setup menu into PCDs allowing for easy customisation.
Added a Dasharo theme as a bonus and PoC.

image

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested
qemu-system-x86_64 -machine q35 -serial telnet:localhost:1345,server,nowait -drive if=pflash,format=raw,unit=0,file=build/coreboot.rom

### DIFF
--- a/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
+++ b/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
@@ -55,6 +55,7 @@ gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPickListBackgroundColor|0x30       # Wh
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowTextColor|0x0F                # White text
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowBackgroundColor|0x60          # Grey bg
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldGrayedTestColor|0x07          # Light grey text
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserErrorTextColor|0x0C                # Bright Red text
 
 gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlack      | { 0x00, 0x00, 0x00, 0x00 }    # 0 Plain black
 gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlue  | { 0x00, 0x00, 0x00, 0x98 }    # 1 Dark green
--- a/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
+++ b/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
@@ -1,4 +1,4 @@
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldBackgroundHighlightColor|0x60 # 
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldBackgroundHighlightColor|0x60 
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldTextHighlightColor|0x0F
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserTitleBackgroundColor|0x60
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserTitleTextColor|0x0F
@@ -23,20 +23,20 @@ gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowTextColor|0x0F
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowBackgroundColor|0x60
 gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldGrayedTestColor|0x07
 
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlack      | { 0x00, 0x00, 0x00, 0x00 }    # Plain black
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightBlue  | { 0x00, 0x00, 0x00, 0x98 }    # Dark green
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGreen | { 0x00, 0x38, 0xD4, 0x30 }    # Dasharo green
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightCyan  | { 0x00, 0xFF, 0xFF, 0xFF }    # White
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightRed   | { 0x00, 0x98, 0x00, 0x00 }    # Red
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteMagenta    | { 0x00, 0x98, 0x00, 0x98 }    # Magenta
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBrown      | { 0x00, 0x29, 0x36, 0x3b }    # Dasharo grey
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGray  | { 0x00, 0x80, 0x80, 0x80 }    # Light grey
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlack      | { 0x00, 0x00, 0x00, 0x00 }    # 0 Plain black
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightBlue  | { 0x00, 0x00, 0x00, 0x98 }    # 1 Dark green
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGreen | { 0x00, 0x38, 0xD4, 0x30 }    # 2 Dasharo green
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightCyan  | { 0x00, 0xFF, 0xFF, 0xFF }    # 3 White
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightRed   | { 0x00, 0x98, 0x00, 0x00 }    # 4 Red
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteMagenta    | { 0x00, 0x98, 0x00, 0x98 }    # 5 Magenta
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBrown      | { 0x00, 0x29, 0x36, 0x3b }    # 6 Dasharo grey
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGray  | { 0x00, 0x80, 0x80, 0x80 }    # 7 Light grey
 
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteDarkGray   | { 0x00, 0x30, 0x30, 0x30 }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlue       | { 0x00, 0x00, 0x00, 0xFF }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLime       | { 0x00, 0x00, 0xFF, 0x00 }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteCyan       | { 0x00, 0xFF, 0xFF, 0x00 }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteRed        | { 0x00, 0xFF, 0x00, 0x00 }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteFuchsia    | { 0x00, 0xFF, 0x00, 0xFF }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteYellow     | { 0x00, 0xFF, 0xFF, 0x00 }
-gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteWhite      | { 0x00, 0xFF, 0xFF, 0xFF }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteDarkGray   | { 0x00, 0x30, 0x30, 0x30 }    # 00 Dark grey
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlue       | { 0x00, 0x00, 0x00, 0xFF }    # 10 Blue
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLime       | { 0x00, 0x00, 0xFF, 0x00 }    # 20 Green
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteCyan       | { 0x00, 0xFF, 0xFF, 0x00 }    # 30 Cyan
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteRed        | { 0x00, 0xFF, 0x00, 0x00 }    # 40 Red
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteFuchsia    | { 0x00, 0xFF, 0x00, 0xFF }    # 50 Fuchsia
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteYellow     | { 0x00, 0xFF, 0xFF, 0x00 }    # 60 Yellow
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteWhite      | { 0x00, 0xFF, 0xFF, 0xFF }    # 70 White
--- a/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
+++ b/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
@@ -1,0 +1,42 @@
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldBackgroundHighlightColor|0x60 # 
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldTextHighlightColor|0x0F
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserTitleBackgroundColor|0x60
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserTitleTextColor|0x0F
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupBackgroundColor|0x60
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupTextColor|0x0F
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupInverseTextColor|0x30
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupInverseBackgroundColor|0x06
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserHelpTextColor|0x07
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserHelpBackgroundColor|0x30
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserInfoTextColor|0x06
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserInfoBackgroundColor|0x30
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldTextColor|0x06
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldBackgroundColor|0x30
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserSubtitleTextColor|0x02
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserBannerBackgroundColor|0x60
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserBannerTextColor|0x0F
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserKeyHelpTextColor|0x02
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserKeyHelpBackgroundColor|0x60
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPickListTextColor|0x06
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPickListBackgroundColor|0x30
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowTextColor|0x0F
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowBackgroundColor|0x60
+gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldGrayedTestColor|0x07
+
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlack      | { 0x00, 0x00, 0x00, 0x00 }    # Plain black
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightBlue  | { 0x00, 0x00, 0x00, 0x98 }    # Dark green
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGreen | { 0x00, 0x38, 0xD4, 0x30 }    # Dasharo green
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightCyan  | { 0x00, 0xFF, 0xFF, 0xFF }    # White
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightRed   | { 0x00, 0x98, 0x00, 0x00 }    # Red
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteMagenta    | { 0x00, 0x98, 0x00, 0x98 }    # Magenta
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBrown      | { 0x00, 0x29, 0x36, 0x3b }    # Dasharo grey
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGray  | { 0x00, 0x80, 0x80, 0x80 }    # Light grey
+
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteDarkGray   | { 0x00, 0x30, 0x30, 0x30 }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlue       | { 0x00, 0x00, 0x00, 0xFF }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLime       | { 0x00, 0x00, 0xFF, 0x00 }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteCyan       | { 0x00, 0xFF, 0xFF, 0x00 }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteRed        | { 0x00, 0xFF, 0x00, 0x00 }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteFuchsia    | { 0x00, 0xFF, 0x00, 0xFF }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteYellow     | { 0x00, 0xFF, 0xFF, 0x00 }
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteWhite      | { 0x00, 0xFF, 0xFF, 0xFF }
--- a/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
+++ b/DasharoPayloadPkg/DasharoColorTheme.dsc.inc
@@ -1,27 +1,55 @@
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldBackgroundHighlightColor|0x60 
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldTextHighlightColor|0x0F
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserTitleBackgroundColor|0x60
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserTitleTextColor|0x0F
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupBackgroundColor|0x60
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupTextColor|0x0F
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupInverseTextColor|0x30
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupInverseBackgroundColor|0x06
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserHelpTextColor|0x07
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserHelpBackgroundColor|0x30
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserInfoTextColor|0x06
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserInfoBackgroundColor|0x30
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldTextColor|0x06
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserFieldBackgroundColor|0x30
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserSubtitleTextColor|0x02
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserBannerBackgroundColor|0x60
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserBannerTextColor|0x0F
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserKeyHelpTextColor|0x02
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserKeyHelpBackgroundColor|0x60
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPickListTextColor|0x06
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPickListBackgroundColor|0x30
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowTextColor|0x0F
-gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserArrowBackgroundColor|0x60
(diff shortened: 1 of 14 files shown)


miczyg1:
I just wondered recently when someone will dare to touch CustomizedDisplayLib and play with it :)

We would have to somehow match the Dasharo green colour (on dasharo.com page the hex color for green is `#38d430`). Of course white and black should also be present.

If I color pick the green from the picture you posted, it gives me `#009800` which is not the right one.

philipanda:
> I just wondered recently when someone will dare to touch CustomizedDisplayLib and play with it :)
> 
> We would have to somehow match the Dasharo green colour (on dasharo.com page the hex color for green is `#38d430`). Of course white and black should also be present.
> 
> If I color pick the green from the picture you posted, it gives me `#009800` which is not the right one.

CustomizedDisplayLib only supports a small color palette which looks similar to VGA color palette. There are a total of 8 possible background colors and 16 text colors, so sadly there is little possibilities to play with.

There are only two green shades possible for text - green and light green, for backgrounds its only the green you see.

I can't tell if it would be feasible to extend the color palette. It might be more challenging than just exposing the colors as PSDs.

miczyg1:
I think it will be inevitable to change `mGraphicsEfiColors` table in MdeModulePkg/Universal/Console/GraphicsConsoleDxe/GraphicsConsole.c 

Also we have to take care of serial terminal colors in `TerminalConOutSetAttribute` (MdeModulePkg/Universal/Console/TerminalDxe/TerminalConOut.c)

Also you may try to change the font to something more close to what we use on dasharo.com (I believe it is oswald font regular-400 or a Font Awesome 6 regular-400)

miczyg1:
Colors taken from dasharo.com that I suggest that may be suitable:

DARKGRAY - `#29363b`
LIGHTGRAY - `#808080`
LIGHTGREEN - `#38d430`

philipanda:
I've exposed the hex color palette as PCDs too and come up with this color scheme in graphics mode using the hex values you've given:
image

There are only 8 possible background colors, so I had to do some non-obvious "reshades":
- EFI_LIGHTCYAN -> `#FFFFFF`
- EFI_LIGHTGREEN ->  `#38d430`
- EFI_BROWN ->  `#29363b`
- EFI_LIGHTGRAY -> `#808080`

For now it doesn't look that great via serial though ;)
image

miczyg1:
> For now it doesn't look that great via serial though ;)

Of course it doesn't :) Check out this: https://gist.github.com/viniciusdaniel/53a98cbb1d8cac1bb473da23f5708836
Maybe the 88/256 colors are supported. If not, then it would have to be implemented in `TerminalConOutSetAttribute`
Currently I believe we use 8/16 colors only as they are widely supported. The 88/256 palette seem to be supported by just a handful of terminals. There is also 24bit color: https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit (then maybe the same graphics RGB notations could be used in serial terminal)

miczyg1:
Couple more suggestions to the color selection:

1. Option text (Intel ME mode) and simple text (Press ESC to exit) colors maybe should be reversed? Options green and text gray. Help text is good though.
2. Originally the hotkey help section had a black foreground to stand out a bit differently. Can it stay black? It would probably look good with green hotkey help strings.
3. Help section should be darker. It barely stands out from the foreground:

image

DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -78,7 +78,7 @@
   # Vendor FFFF Device 0000 Prog Interface 1, BAR #0, Offset 0, Stride = 1, Clock 1843200 (0x1c2000)
   #
   #                                       [Vendor]   [Device]  [----ClockRate---]  [------------Offset-----------] [Bar] [Stride] [RxFifo] [TxFifo]   [Rsvd]   [Vendor]
-  DEFINE PCI_SERIAL_PARAMETERS        = {0xff,0xff, 0x00,0x00, 0x0,0x20,0x1c,0x00, 0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0, 0x00,    0x01, 0x0,0x0, 0x0,0x0, 0x0,0x0, 0xff,0xff}
+  DEFINE PCI_SERIAL_PARAMETERS        = {0xff,0xff, 0x00,0x00, 0x0,0x60,0x1c,0x00, 0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0, 0x00,    0x01, 0x0,0x0, 0x0,0x0, 0x0,0x0, 0xff,0xff}
```
miczyg1: Why change this?

DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -674,6 +674,49 @@
...
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGray  | { 0x00, 0x80, 0x80, 0x80 }
+
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteDarkGray   | { 0x00, 0x30, 0x30, 0x30 }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteBlue       | { 0x00, 0x00, 0x00, 0xFF }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLime       | { 0x00, 0x00, 0xFF, 0x00 }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteCyan       | { 0x00, 0xFF, 0xFF, 0x00 }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteRed        | { 0x00, 0xFF, 0x00, 0x00 }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteFuchsia    | { 0x00, 0xFF, 0x00, 0xFF }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteYellow     | { 0x00, 0xFF, 0xFF, 0x00 }
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteWhite      | { 0x00, 0xFF, 0xFF, 0xFF }
```
miczyg1: This could be moved to a file like `DasharoPayloadPkg/DasharoColorStyle.dsc.inc`, theen it could be quickly included in other places in the repo (e.g. OVMF)

Also some comments at the end of lines what color does given PcdBrowser* correspond to, would be great

philipanda:
> I think it will be inevitable to change `mGraphicsEfiColors` table in MdeModulePkg/Universal/Console/GraphicsConsoleDxe/GraphicsConsole.c
> 
> Also we have to take care of serial terminal colors in `TerminalConOutSetAttribute` (MdeModulePkg/Universal/Console/TerminalDxe/TerminalConOut.c)
> 
> Also you may try to change the font to something more close to what we use on dasharo.com (I believe it is oswald font regular-400 or a Font Awesome 6 regular-400)

That might be more troublesome than expected :laughing: 
It seems that the font used by GraphicsConsole.c is defined at `/MdeModulePkg/Universal/Console/GraphicsConsoleDxe/LaffStd.c`. That is a huge array that maps Unicode glyphs to bitmaps made of hex bytes.

Adding a support for vector fonts might be a little too much for now, but maybe it could be feasible to create a tool that could convert vector fonts into bitmaps like the one defined in `LaffStd.c`.

philipanda:
I've been fiddling around with converting the Oswald font to bitmaps. The results seemed promising, but in the end I think it will not be possible to properly display a proper vector font in 19x8 resolution. 

That's the best results I was able to achieve. While it's possible to do some postprocessing to fix jagged edges, losing detail is inevitable. Creating such a small font requires working with pixel art.

image

The scripts for the conversion are [on a separate branch](https://github.com/Dasharo/edk2/tree/ttf-to-bitmap-fonts) as I will scrap that idea for now.

macpijan:
Yeah the font does not look great and would require more work.

The colors look interesting, though.

DasharoPayloadPkg/DasharoPayloadPkg.dsc
philipanda: https://github.com/Dasharo/edk2/pull/263/commits/145ff2b6be3c1f85e76499282ed431789c73b0e8
https://github.com/Dasharo/edk2/pull/263/commits/ed696ff6b613ece43ff13faae63a192408f009f8

DasharoPayloadPkg/DasharoPayloadPkg.dsc
philipanda: It's was a search&replace artifact, thanks for noticing it
https://github.com/Dasharo/edk2/pull/263/commits/145ff2b6be3c1f85e76499282ed431789c73b0e8#diff-4c9d32b9f7d7dfb40b295a05a40ad511e85a6d9e499e7901abc79d21f9b18833L81

philipanda:
Looks like the 256 colors mode might work easily on terminals that support it.
It's enough to modify the escape codes produced by the `TerminalConOutSetAttribute`. I don't think there would be a simple way of knowing what color modes the client terminal understands though, so the compatibility with simple terminals might get worse.

I've tried hardcoding this pinkish color for every printed character. I think I've messed something up with the escape codes, but the color definitely looks like the one I've requested (213).
image

philipanda:
And the 24 bit mode works perfectly too, at least on Gnome terminal.
The amount of work that would be needed to implement any of them seem to be similarly low. It would require only modifying how `TerminalConOutSetAttribute` encodes the 16 color EFI color palette into escape codes.

The color used here is the Dasharo green #38d430.
The only little issue is that the color needs to be convered from hex to decimal to use in an escape code.
image

When accessing the serial without the 24b truecolor support using xterm:
```
xterm -cm
```
Nothing is broken, just that there are no colors:
image

I think 24b color would be the best solution. I had a hard time trying to stop my terminal from interpreting them, so that makes me believe that most potential users won't have issues with that and it won't break tests via serial in OSFV.

philipanda:
Using the 24b escape codes the colors work great via on a serial console via telnet on Qemu:
image

Some of the colors, like most of the gray/white text, don't exactly match. I suspect it's an issue with mapping the RGB colors into the 16 EFI colors along the way. Maybe the `bright` bit is not interpreted correctly when mapping the colors.

miczyg1:
> The only little issue is that the color needs to be convered from hex to decimal to use in an escape code.

Should be easy with `%02u` formatter for printf.

> Some of the colors, like most of the gray/white text, don't exactly match. I suspect it's an issue with mapping the RGB colors into the 16 EFI colors along the way. Maybe the bright bit is not interpreted correctly when mapping the colors.

Yeah, the letters in the top section are far from being white... But hey, it looks neat already.

DasharoPayloadPkg/vbt.bin
miczyg1, line 1: @philipanda another mistakenly  added change?

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConOut.c
```diff
@@ -562,96 +561,17 @@ TerminalConOutSetAttribute (
...
-  mSetAttributeString[BRIGHT_CONTROL_OFFSET]         = (CHAR16)('0' + BrightControl);
-  mSetAttributeString[FOREGROUND_CONTROL_OFFSET + 0] = (CHAR16)('0' + (ForegroundControl / 10));
-  mSetAttributeString[FOREGROUND_CONTROL_OFFSET + 1] = (CHAR16)('0' + (ForegroundControl % 10));
-  mSetAttributeString[BACKGROUND_CONTROL_OFFSET + 0] = (CHAR16)('0' + (BackgroundControl / 10));
-  mSetAttributeString[BACKGROUND_CONTROL_OFFSET + 1] = (CHAR16)('0' + (BackgroundControl % 10));
+  CHAR16 mSetTerminalColorString[64];
+  UnicodeSPrint(mSetTerminalColorString, sizeof(mSetTerminalColorString), 
+    L"%c[38;2;%u;%u;%um%c[48;2;%u;%u;%um",
+    ESC, ForegroundControl.r, ForegroundControl.g, ForegroundControl.b,
+    ESC, BackgroundControl.r, BackgroundControl.g, BackgroundControl.b);
```
miczyg1:
```
  mSetAttributeString[BRIGHT_CONTROL_OFFSET]         = (CHAR16)('0' + BrightControl);
  mSetAttributeString[FOREGROUND_CONTROL_OFFSET + 0] = (CHAR16)('0' + (ForegroundControl / 10));
  mSetAttributeString[FOREGROUND_CONTROL_OFFSET + 1] = (CHAR16)('0' + (ForegroundControl % 10));
  mSetAttributeString[BACKGROUND_CONTROL_OFFSET + 0] = (CHAR16)('0' + (BackgroundControl / 10));
  mSetAttributeString[BACKGROUND_CONTROL_OFFSET + 1] = (CHAR16)('0' + (BackgroundControl % 10));
```

This approach would work too, but you would have to change the mSetAttributeString string and adjust BACKGROUND_CONTROL_OFFSET, BRIGHT_CONTROL_OFFSET.

DasharoPayloadPkg/vbt.bin
philipanda: Thanks for pointing it out. I've commited some random files before too. I think that I'm commiting some artifacts from the build process that I am unaware of. I certainly did not search&replaced something in a binary file :laughing:. I'll check twice before adding a whole directory next time.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConOut.c
philipanda: I thought it would be more troublesome, because the color code is not one character long this time. The offsets will not be constant. Unless the color codes can have leading zeros, which I haven't checked. Then you would be right. Need to check that, thanks

miczyg1: I see. If the length of the color code is not fixed, then it makes sense for the implementation to change.

philipanda: With RGB color encodings there is much more data to set and I doubt it's worth keeping the current approach. Keeping this implementation with offsets will produce much larger code. In my opinion switching to a formatted sprint is better (defines etc next to each other for visualization in the snippet)

```c
CHAR16  mSetAttributeString[] = {
  ESC,
  '[', '3', '8', ';', '2', ';',
  '0', '0', '0', ';', // foreground R
  '0', '0', '0', ';', // foreground G
  '0', '0', '0', 'm', // foreground B
  ESC,
  '[', '4', '8', ';', '2', ';',
  '0', '0', '0', ';', // background R
  '0', '0', '0', ';', // background G
  '0', '0', '0', 'm', // background B
}
  #define FOREGROUND_CONTROL_R_OFFSET 7
  #define FOREGROUND_CONTROL_G_OFFSET 11
  #define FOREGROUND_CONTROL_B_OFFSET 15
  #define BACKGROUND_CONTROL_R_OFFSET 26
  #define BACKGROUND_CONTROL_G_OFFSET 30
  #define BACKGROUND_CONTROL_B_OFFSET 34

  mSetAttributeString[BRIGHT_CONTROL_OFFSET]         = (CHAR16)('0' + BrightControl);

  mSetAttributeString[FOREGROUND_CONTROL_R_OFFSET + 0] = (CHAR16)('0' + (ForegroundControl.r / 100));
  mSetAttributeString[FOREGROUND_CONTROL_R_OFFSET + 1] = (CHAR16)('0' + (ForegroundControl.r % 100 / 10));
  mSetAttributeString[FOREGROUND_CONTROL_R_OFFSET + 2] = (CHAR16)('0' + (ForegroundControl.r % 10));

  mSetAttributeString[FOREGROUND_CONTROL_G_OFFSET + 0] = (CHAR16)('0' + (ForegroundControl.g / 100));
  mSetAttributeString[FOREGROUND_CONTROL_G_OFFSET + 1] = (CHAR16)('0' + (ForegroundControl.g % 100 / 10));
  mSetAttributeString[FOREGROUND_CONTROL_G_OFFSET + 2] = (CHAR16)('0' + (ForegroundControl.g % 10));

  mSetAttributeString[FOREGROUND_CONTROL_B_OFFSET + 0] = (CHAR16)('0' + (ForegroundControl.b / 100));
  mSetAttributeString[FOREGROUND_CONTROL_B_OFFSET + 1] = (CHAR16)('0' + (ForegroundControl.b % 100 / 10));
  mSetAttributeString[FOREGROUND_CONTROL_B_OFFSET + 2] = (CHAR16)('0' + (ForegroundControl.b % 10));

  mSetAttributeString[BACKGROUND_CONTROL_R_OFFSET + 0] = (CHAR16)('0' + (BackgroundControl.r / 100));
  mSetAttributeString[BACKGROUND_CONTROL_R_OFFSET + 1] = (CHAR16)('0' + (BackgroundControl.r % 100 / 10));
  mSetAttributeString[BACKGROUND_CONTROL_R_OFFSET + 2] = (CHAR16)('0' + (BackgroundControl.r % 10));

  mSetAttributeString[BACKGROUND_CONTROL_G_OFFSET + 0] = (CHAR16)('0' + (BackgroundControl.g / 100));
  mSetAttributeString[BACKGROUND_CONTROL_G_OFFSET + 1] = (CHAR16)('0' + (BackgroundControl.g % 100 / 10));
  mSetAttributeString[BACKGROUND_CONTROL_G_OFFSET + 2] = (CHAR16)('0' + (BackgroundControl.g % 10));

  mSetAttributeString[BACKGROUND_CONTROL_B_OFFSET + 0] = (CHAR16)('0' + (BackgroundControl.b / 100));
  mSetAttributeString[BACKGROUND_CONTROL_B_OFFSET + 1] = (CHAR16)('0' + (BackgroundControl.b % 100 / 10));
  mSetAttributeString[BACKGROUND_CONTROL_B_OFFSET + 2] = (CHAR16)('0' + (BackgroundControl.b % 10));
  ```

**Comment by philipanda** [comment]
It looks like whatever exact code prints the graphical version of setup menu correctly uses the code `0x0F` for white foreground, but the serial console uses `0x07` instead (no bright bit). It only affects this single color. Every other shade is correct in both the graphical and text (serial) modes. I guess the value `0x07` has to be hardcoded somewhere and I don't yet know where, or it is not updated witht the `DasharoColorTheme.dsc.inc` as `0x07` would be the default value from EDK2

**Comment by philipanda** [comment]
> It looks like whatever exact code prints the graphical version of setup menu correctly uses the code 0x0F for white foreground, but the serial console uses 0x07 instead (no bright bit). It only affects this single color. Every other shade is correct in both the graphical and text (serial) modes. I guess the value 0x07 has to be hardcoded somewhere and I don't yet know where, or it is not updated witht the DasharoColorTheme.dsc.inc as 0x07 would be the default value from EDK2

The issue was the line `MdeModulePkg/Universal/Console/TerminalDxe/TerminalConOut.c:564` where the bright bit was being cleared: `ForegroundControl = mTerminalEfiColors[Attribute & 0x07];`. Not sure why it was like that before as the foreground supports the bright bit, unlike the background. Anding with `0x0F` is the way to go and fixes the issues with invalid colors on the serial console.

With that out of the way I think all issues with this PR are fixed?

**Comment by philipanda** [comment]
The exact pick of the colors is due to subjective feel. I'd suggest first considering only the implementation without defining custom color palette and adding the custom colors later if the one defined in this PR would need some adjustments.

I'd very much like to try and get that color customization into the upstream EDK2 next and would be grateful for any tips to improve the chances of that happening. From me, I'm afraid it is not a complete feature if only the setup menu can be colored. I've never checked how other places like the popups behave. Maybe that should be worked on first.

Another thing is that this change will make all the terminals that only support the classic 4 bit VGA color escape codes show no colors, like in the xterm example. I can't come up with a realistic scenario where someone would use EDK2 via serial using a terminal that does not support 24b RGB colors escape codes, but maybe some switch or a key combination for restoring the 4b mode would be reasonable.

Summing up how this looks like:

- graphical with serial ON:
   Screenshot From 2025-12-20 20-56-57

- serial with rgb support:
   Screenshot From 2025-12-20 20-58-25

- serial with no colors support:
   Screenshot From 2025-12-20 20-57-11

- graphical with serial redirection OFF:
   Screenshot From 2025-12-20 20-58-53

File: DasharoPayloadPkg/vbt.bin
Comment (philipanda) [reply DasharoPayloadPkg/vbt.bin:1]: https://github.com/Dasharo/edk2/compare/6d902d2a1272bd1fee41bb4f0de059cf21365174..d05eb35f3a99e860b23e2b8f55c5bde286a07a79

**Comment by philipanda** [comment]
*rebased onto the dasharo branch and cleared up the commits a bit

**Comment by pietrushnic** [comment]
Looks cool. I will not judge the implementation, but definitely having our theme is a nice direction. It gives us some branding alignment.

File: DasharoPayloadPkg/DasharoColorTheme.dsc.inc [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGray  | { 0x00, 0x80, 0x80, 0x80 }    # 7 Light grey
+
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteDarkGray   | { 0x00, 0x30, 0x30, 0x30 }    # 8 Dark grey
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightBlue      | { 0x00, 0x00, 0x00, 0xFF }    # 9 Blue
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGreen       | { 0x00, 0x00, 0xFF, 0x00 }    # A Green
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightCyan       | { 0x00, 0xFF, 0xFF, 0x00 }    # B Cyan
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightRed        | { 0x00, 0xFF, 0x00, 0x00 }    # C Red
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteMagenta    | { 0x00, 0xFF, 0x00, 0xFF }    # D Fuchsia
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteYellow     | { 0x00, 0xFF, 0xFF, 0x00 }    # E Yellow
+gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteWhite      | { 0x00, 0xFF, 0xFF, 0xFF }    # F White
```
miczyg1: Needs a newline

MdeModulePkg/Library/CustomizedDisplayLib/CustomizedDisplayLib.c
```diff
@@ -153,7 +153,7 @@ RefreshKeyHelp (
     return;
   }
 
-  gST->ConOut->SetAttribute (gST->ConOut, KEYHELP_TEXT | KEYHELP_BACKGROUND);
```
miczyg1, line 156: I wonder if it would not be easier to replace definitions of these macros in `MdeModulePkg/Library/CustomizedDisplayLib/Colors.h`: like this:

```diff
-#define KEYHELP_TEXT                   EFI_LIGHTGRAY
-#define KEYHELP_BACKGROUND             EFI_BACKGROUND_BLACK
+#define KEYHELP_TEXT                   PcdGet8 (PcdBrowserKeyHelpTextColor) 
+#define KEYHELP_BACKGROUND             PcdGet8 (PcdBrowserKeyHelpBackgroundColor)
```

Then we could remove all those `Get*Color` helpers

MdeModulePkg/Universal/Console/GraphicsConsoleDxe/GraphicsConsole.c
```diff
@@ -60,29 +60,53 @@ EFI_GUID  mFontPackageListGuid = {
...
+
+/**
+  test
+**/
+VOID
+InitGraphicsColors (
+  VOID
+  )
+{
+  CopyGraphicsPaletteColor(&mGraphicsEfiColors[0],  PcdGetPtr(PcdGraphicsConsoleColorPaletteBlack));
```
miczyg1, line 92: I wonder if FixedPcdGetPtr will work like this:

```C
EFI_GRAPHICS_OUTPUT_BLT_PIXEL  mGraphicsEfiColors[16] = {
  //
  // B    G    R   reserved
  //
  FixedPcdGetPtr(PcdGraphicsConsoleColorPaletteBlack)),
  ...
  FixedPcdGetPtr(PcdGraphicsConsoleColorPaletteWhite)),
};
```

Would be much cleaner, if it does work.

miczyg1 (changes requested):

miczyg1:
> I'd very much like to try and get that color customization into the upstream EDK2 next and would be grateful for any tips to improve the chances of that happening. From me, I'm afraid it is not a complete feature if only the setup menu can be colored. I've never checked how other places like the popups behave. Maybe that should be worked on first.

Pop-ups are created in two ways:

1. `CreatePopUp` in MdePkg/Library/UefiLib/Console.c. Attributes (colors) are taken as an argument, so it depends on the application that creates the pop-up which color it should look like. We can't possibly change it. Even if it is `EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE` (like until now, in most cases) I think it is fine. Unless we `sed` all calls to `CreatePopUp` and change the colors...
2. `HiiPopup->CreatePopup` using the `EFI_HII_POPUP_PROTOCOL`. The implementation is in MdeModulePkg/Universal/DisplayEngineDxe/Popup.c and uses the color returned by `GetPopupColor`, which you already modified to pick color from PCDs in the `CustomizedDisplayLib`.

So we would only have to handle case 1. So far, I have seen `EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE` everywhere, except the FUM warning in `DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c`, which uses `EFI_BLACK | EFI_BACKGROUND_RED`.

MdeModulePkg/Universal/Console/GraphicsConsoleDxe/GraphicsConsole.c
philipanda: Is `FixedPcdGetPtr` supposed to be accessible in compile time, as opposed to `PcdGetPtr`? Then it should compile if we add pointer casts and a dereference.

Technically it could be possible to cast an arbitrary buffer pointer to a `EFI_GRAPHICS_OUTPUT_BLT_PIXEL` pointer and dereference it. Unless the compiler decides to add some padding or change the order inside the struct. While mechanisms like `__attribute__((packed))` exist, I think it might be a bit too hacky while not being strictly necessary. What do you think?

philipanda:
> So we would only have to handle case 1. So far, I have seen EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE everywhere, except the FUM warning in DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c, which uses EFI_BLACK | EFI_BACKGROUND_RED.

If `EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE` is used almost everywhere, then a new set of two PCDs could be created to represent a default popup color. For other cases, like the FUM, it should be up to the creator of the app to either hardcode the colors, use existing, or create new PCDs. In this case it shouldn't be a problem to add a PCD for the unique FUM popup to the color palette configs.

MdeModulePkg/Universal/Console/GraphicsConsoleDxe/GraphicsConsole.c
miczyg1: I think I am mixing things here. mGraphicsEfiColors is a table of `EFI_GRAPHICS_OUTPUT_BLT_PIXEL`s not a table of `EFI_GRAPHICS_OUTPUT_BLT_PIXEL *` (pointers), so that won't work. Unless we would rewrite the code to dereference pointers from the table... Not sure though if `FixedPcdGetPtr` would work here as I thought. Closing this thread.

miczyg1:
> > So we would only have to handle case 1. So far, I have seen EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE everywhere, except the FUM warning in DasharoModulePkg/Library/DasharoSystemFeaturesUiLib/DasharoSystemFeatures.c, which uses EFI_BLACK | EFI_BACKGROUND_RED.
> 
> If `EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE` is used almost everywhere, then a new set of two PCDs could be created to represent a default popup color. For other cases, like the FUM, it should be up to the creator of the app to either hardcode the colors, use existing, or create new PCDs. In this case it shouldn't be a problem to add a PCD for the unique FUM popup to the color palette configs.

We already have the PCDs:  `gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupBackgroundColor`, 
`gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupTextColor` that are responsible for it. But the problem is, that we would have to rewrite a significant amount of code for these popups if we would like to use these.

I think the best we can do is to replace all `EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE` with a new macro `EFI_DEFAULT_POPUP_COLOR` which would be defined as  0xFFFFFFFF (UINT32_MAX).  The definition of `CreatePopUp` in UefiLib takes the attribute as UINTN (32bit or 64bit value depending on build/architecture), so we have more bits than 8 at disposal. So, if the Attribute input parameter would be equal to 0xFFFFFFFF (`EFI_DEFAULT_POPUP_COLOR`), then we will use `gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupBackgroundColor` and `gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupTextColor`, otherwise use the colors passed via the Attribute parameter. That way we only have to add these tw PCDs to UefiLib and nothing else.

philipanda:
> We already have the PCDs: gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupBackgroundColor,
gEfiMdeModulePkgTokenSpaceGuid.PcdBrowserPopupTextColor that are responsible for it

Right, I thought they were used for some other popups I've never seen before looking that the colors are not `EFI_LIGHTGRAY | EFI_BACKGROUND_BLUE`, but I've looked on the Dasharo palette instead of the default values, so you're right, we already have these PCDs.

MdeModulePkg/Library/CustomizedDisplayLib/CustomizedDisplayLib.c
miczyg1: @philipanda do you plan to change it per suggestion?

DasharoPayloadPkg/DasharoColorTheme.dsc.inc
miczyg1: @philipanda the newline is still missing

MdeModulePkg/Library/CustomizedDisplayLib/CustomizedDisplayLib.c
philipanda: Sorry, I didn't have much time to work on the PR lately.
Your suggestion makes sense, I do plan to change that

miczyg1: I saw that you rebased it, so I assumed you have found some time :)

DasharoPayloadPkg/DasharoColorTheme.dsc.inc
philipanda: https://github.com/Dasharo/edk2/pull/263/changes/166e673ebffca722396835ae48fd16ae5208cb5d

MdeModulePkg/Library/CustomizedDisplayLib/CustomizedDisplayLib.c
philipanda: After looking into it deeper I start to doubt removing the helpers and using the defines instead will make it simpler or cleaner.
There were already a couple of such helpers available because some color mixes were used multiple times across the code, example `GetFieldTextColor`.

I think the cleanest solution would be to make a macro like:
```c
#define GetTextColor(type) (type##Text | type##Background)
```
and go with defining every foreground/background combination in the Pcds and remove all the helpers.

Otherwise it will stay a mess with multiple ways of accessing them, because some places would use (define_f | define_b) and others would call the helpers.

If so then all the usages would need to be updated too.
I think it would be a good change, but I'm not sure such a cleanup should be bundled along this PR?

philipanda: https://github.com/Dasharo/edk2/pull/263/commits/9f4ccf624bd0fbc8ef1c34fd3631be347b683f2a applies that to only the helpers that were added in this PR before.
Added missing ErrorText color PCD too 
image

miczyg1: > I think it would be a good change, but I'm not sure such a cleanup should be bundled along this PR?

Well, you are redesigning the whole color styles. If not here, then where? Can you assure that nothing will break if you skip it in this PR?

MdeModulePkg/Universal/Console/TerminalDxe/Terminal.c
```diff
@@ -1409,3 +1411,33 @@ IsHotPlugDevice (
 
   return FALSE;
 }
+
+TerminalEfiColor mTerminalEfiColors[16];
+
+VOID CopyTerminalPaletteColor(IN UINT8* Src, OUT TerminalEfiColor* Dest) {
```
miczyg1, line 1417: Usual order of parameters for copy methods is `Copy(Dest, Src)`

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -2001,6 +2001,48 @@
...
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGray  | { 0x00, 0x98, 0x98, 0x98 } | VOID* | 0x00010407
+  
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteDarkGray   | { 0x00, 0x30, 0x30, 0x30 } | VOID* | 0x00010408
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightBlue      | { 0x00, 0x00, 0x00, 0xFF } | VOID* | 0x00010409
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightGreen       | { 0x00, 0x00, 0xFF, 0x00 } | VOID* | 0x0001040A
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightCyan       | { 0x00, 0xFF, 0xFF, 0x00 } | VOID* | 0x0001040B
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightRed        | { 0x00, 0xFF, 0x00, 0x00 } | VOID* | 0x0001040C
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteMagenta    | { 0x00, 0xFF, 0x00, 0xFF } | VOID* | 0x0001040D
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteYellow     | { 0x00, 0xFF, 0xFF, 0x00 } | VOID* | 0x0001040E
+  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteWhite      | { 0x00, 0xFF, 0xFF, 0xFF } | VOID* | 0x0001040F
```
miczyg1, line 2042: Please remove the redundant spaces. They are not even aligned properly for readability

philipanda: I thought they helped with readibility by splitting it into two blocks of 8 each.
Can remove, sure

MdeModulePkg/Library/CustomizedDisplayLib/CustomizedDisplayLib.c
philipanda: I am confident nothing will break if I don't touch these public functions. I won't be so sure nothing breaks in other modules if I do. If some automatic validation would be enough to ensure it, that would be good. I'm just a bit worried about the size and duration of this PR. If it's preferred to deliver full feature-complete changes, then I guess it cannot be helped :)

MdeModulePkg/MdeModulePkg.dec
miczyg1: I meant the spaces near the pipes  `|`, e.g.

```
  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteLightRed        | { 0x00, 0xFF, 0x00, 0x00 } | VOID* | 0x0001040C
  gEfiMdeModulePkgTokenSpaceGuid.PcdGraphicsConsoleColorPaletteMagenta    | { 0x00, 0xFF, 0x00, 0xFF } | VOID* | 0x0001040D
```

See? Eveywhere else in the repo there is no space around pipe character `|`, so I would do the same here.

philipanda: right, I see now :+1:

PR #271, opened 2025-08-28
DasharoPayloadPkg: recover variable storage damaged by a reboot

### LINKED ISSUE #1293: V54: BIOS settings are randomly reset
### Component

Dasharo firmware

### Device

NovaCustom V54 14th Gen

### Dasharo version

v0.9.0

### Dasharo Tools Suite version

_No response_

### Test case ID

_No response_

### Brief summary

With a new V54x_6x_TU laptop, it happened a few times now, that we recognized a reset of some important BIOS settings like secure boot, camera disable, Intel ME etc. Currently, we can't confirm that every setting got reset. We have noticed it on Ubuntu 24.04 in the security audit settings that secure boot was suddenly disabled after working with it. We can't trust the BIOS settings anymore because of the unstable state.
There is definitely no long startup, which would indicate a memory re-training.
Did anyone observe something like this?

### How reproducible

Not reproducible / happens randomly

### How to reproduce

We can't reproduce it currently.

### Expected behavior

BIOS settings should be persistent.

### Actual behavior

BIOS settings like secure boot or Intel ME are reset.

### Screenshots

_No response_

### Additional context

_No response_

### Solutions you've tried

* CMOS battery connector is stable and fit correctly
* CMOS battery has stable voltage, checked with voltmeter
* Reset CMOS battery according to Novacustom; After that, laptop needs more time as expected to re-train the memory

SergiiDmytruk:
FaultTolerateWrite is enabled and employed, but no recovery was performed.  Not even sure how it happens without coreboot, there is something in MdeModulePkg/Universal/Variable/Pei and MdeModulePkg/Universal/Variable/Pei, but operation of tha code isn't obvious.  Either way, it doesn't kick in for coreboot.

Here is how it works:
 1. SmmStorePei uses data from CBMEM to create gVariableFlashInfoHobGuid HOB
 2. FaultTolerantWritePei consumes gVariableFlashInfoHobGuid through VariableFlashInfoLib and produces gEdkiiFaultTolerantWriteGuid
 3. BlSupportPei consumes gEdkiiFaultTolerantWriteGuid and uses it to decide which boot mode to use
 4. When SmmStoreFvbRuntime gets initialized it looks up gEdkiiFaultTolerantWriteGuid and if needed finishes the operation interrupted by a reboot

coreboot part: https://github.com/Dasharo/coreboot/pull/758
fixes: https://github.com/Dasharo/dasharo-issues/issues/1293
*ref: ncm-1969*

### DIFF
--- a/DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
+++ b/DasharoPayloadPkg/BlSupportPei/BlSupportPei.c
@@ -814,7 +814,12 @@ BlPeiEntryPoint (
     DEBUG ((DEBUG_INFO, "Created SMMSTORE info hob\n"));
 
     Status = ValidateFvHeader (&SMMSTOREInfo);
-    if (EFI_ERROR (Status)) {
+    //
+    // gEdkiiFaultTolerantWriteGuid HOB doesn't exist if both working and spare
+    // parts of the variable store are invalid, this is when we need to
+    // initialize it with defaults.
+    //
+    if (EFI_ERROR (Status) && GetFirstGuidHob (&gEdkiiFaultTolerantWriteGuid) == NULL) {
       Status = PeiServicesSetBootMode (BOOT_WITH_DEFAULT_SETTINGS);
       DEBUG ((DEBUG_INFO, "BootMode: Boot with default settings\n"));
       ASSERT_EFI_ERROR (Status);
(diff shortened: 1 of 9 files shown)


miczyg1:
Oh my. I thought the recovery worked, up till now. I guess insufficient understanding of FVB-related protocols/modules on our side. Good catch.

filipleple (approved):
It's been tested and it works

PR #272, opened 2025-08-29
DMP/Library/DasharoSystemFeaturesUiLib: Workaround IBG_STATUS structu…

miczyg1:


miczyg1:
```c
typedef union {
  ACM_STATUS_BITS Bits;
  UINT64 Raw;
} ACM_STATUS;
```

The `ACM_STATUS_BITS` is 32bit, so the union should probably have 32bit raw value. But I can't see how it would fix the issue.

mkopec (approved):

PR #274, opened 2025-09-01
Descriptor lockdown

mkopec:
Needs https://github.com/Dasharo/coreboot/pull/760

### DIFF
--- a/DasharoModulePkg/Include/DasharoOptions.h
+++ b/DasharoModulePkg/Include/DasharoOptions.h
@@ -44,7 +44,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define DASHARO_VAR_CORE_ACTIVE_COUNT             L"CoreActiveCount"
 #define DASHARO_VAR_HYPER_THREADING               L"HyperThreading"
 #define DASHARO_VAR_USB_PORT_POWER                L"UsbPortPower"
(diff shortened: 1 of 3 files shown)


mkopec:
Does not work yet, it seems that reading the DescriptorWriteable EFI variable fails.

mkopec:
Tested: Booted V540TU to the setup menu and verified that the HAP option does not appear anymore - except on first boot, where coreboot cannot set the variable due to the EFI store not being initialized.

Managed to succesfuly initiate Capsule Update twice without setting the ME to HAP mode, too, however the actual update got stuck at a black screen both times and the laptop got bricked. @SergiiDmytruk do you have some ideas?

SergiiDmytruk:
Assuming SMMSTORE wasn't damaged by coreboot, EDK should start executing either way (like it did before HAP check was introduced).  If the screen is black, nothing should have been read from the flash yet, so probably not a hang due to flash not being readable.  The behaviour may be reproducible on other hardware with logging or even in QEMU.  I also don't remember there being checks in other places.

mkopec:
Ahh now that the ME is not in HAP mode, the code tries to enable HMRFPO, which causes a global reset and corrupts the capsule data... Need to remove the HMRFPO enabling code at least until capsule on disk is introduced

macpijan (approved):

SergiiDmytruk (approved):

filipleple (approved):

PR #277, opened 2025-09-17
DasharoPayloadPkg/PlatformBootManagerLib: register serial console after VGA

SergiiDmytruk:
At least some bootloaders (e.g., one in FreeBSD) use the first console of ConOut as the primary one. Putting serial console first leads to interactive output going to serial where most users won't even see it.

OVMF and likely other firmware built with EDK put graphical output before the serial, so this shouldn't cause any breakage although some changes in behaviour is possible.

***

Tagging @miczyg1 in case he had a specific reason for adding consoles in this order, although judging by the comment serial was registered when it could be discovered.

***

This addresses  https://github.com/Dasharo/dasharo-issues/issues/1545.
*ref: prot-1908*


miczyg1 (approved):
As far as I am concerned, the order should not matter. Pretty weird for FreeBSD system to do such a thing. I always thought that all console outputs in ConOut are treated equally.

PR #282, opened 2025-10-30
TPM multibank detection support

filipleple:
coreboot: https://github.com/Dasharo/coreboot/pull/808

_ref: PROT-1960_


DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+      );
+
+      Status = gBS->CloseEvent(TimerEvent);
+      ASSERT_EFI_ERROR(Status);
+      gST->ConOut->EnableCursor(gST->ConOut, CursorVisible);
+      gST->ConOut->SetAttribute(gST->ConOut, CurrentAttribute);
+      gST->ConOut->ClearScreen(gST->ConOut);
+      DrainInput();
+
+      Tcg2PhysicalPresenceLibProcessRequest(NULL);
```
filipleple, line 1222: The way this works on UX side is that after selecting a bank in the `Multiple PCR banks...` popup, you once again get the screen where you confirm the bank change with `F12`.

It kinda feels redundant, but AFAIK this is the way to do this using functions marked as `EFIAPI`, not reimplementing/exporting any static functions. I guess changing PCR banks is not something you do on a daily basis, so it doesn't have to be as snappy as possible

SecurityPkg/Library/DxeTcg2PhysicalPresenceLib/DxeTcg2PhysicalPresenceLib.c
```diff
@@ -193,6 +195,19 @@ Tcg2ExecutePhysicalPresence (
...
       }
 
+      // 
+      // Store the requested value of the active PCR banks for comparison after
+      // reboot, to confirm whether the TPM accepted the change
+      // 
+      gRT->SetVariable(
+        L"RequestedActivePcrBanks",
+        &gEfiTcg2PhysicalPresenceGuid,
+        EFI_VARIABLE_NON_VOLATILE | EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS,
```
SergiiDmytruk: Is there a reason for specifying `EFI_VARIABLE_RUNTIME_ACCESS` for `RequestedActivePcrBanks`?  It seems no OS will need to know about this variable.

SecurityPkg/Library/DxeTcg2PhysicalPresenceLib/DxeTcg2PhysicalPresenceLib.c
```diff
@@ -193,6 +195,19 @@ Tcg2ExecutePhysicalPresence (
         return TCG_PP_OPERATION_RESPONSE_BIOS_FAILURE;
       }
 
+      //
```
SergiiDmytruk, line 197: Did you omit `case TCG2_PHYSICAL_PRESENCE_LOG_ALL_DIGESTS` intentionally?  That operation seems to select all digests and somebody may try using it.

SecurityPkg/Include/Guid/Tcg2PhysicalPresenceData.h
```diff
@@ -35,6 +35,13 @@ typedef struct {
   UINT32    PPFlags;
 } EFI_TCG2_PHYSICAL_PRESENCE_FLAGS;
 
+//
+// This variable is used to store the current active PCRs bitmask, to then
+// compare against it after reboot. This allows verifying if PCR bank change
+// was effective. Zero by default.
+//
+#define REQUESTED_ACTIVE_PCR_BANKS_VARIABLE_NAME  L"RequestedActivePcrBanks"
```
SergiiDmytruk, line 43: This macro gets used only once and `L"RequestedActivePcrBanks"` is used 7 times.  I like the macro more since it has a comment explaining what this variable is for.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
   BootLogoEnableLogo ();
 }
 
+/**
+  Initializes the RequestedActivePcrBanks variable
+**/
+STATIC
+VOID
+EnsurePrevActivePcrBanksVar (
```
SergiiDmytruk: Why not use `if (Status == EFI_NOT_FOUND)` in place of `if (RequestedActivePcrBanks == 0)` in `WarnIfSinglePCRBank()` and just drop this function?

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+  UINT32         RequestedActivePcrBanks;
+  UINTN          Size;
+  UINTN          i;
+  UINTN          OptionCount = 0;
+  UINTN          AvlIdx[ARRAY_SIZE(mHashAlgs)] = {0};
+  CHAR16         OptLine[81];
+  UINT32         SelectedMask;
+
+  Status = Tpm2GetCapabilitySupportedAndActivePcrs(&TpmHashAlgorithmBitmap, &ActivePcrBanks);
+  ASSERT_EFI_ERROR(Status);
```
SergiiDmytruk: So, brick the machine if there is no TPM2? :)  In practice `ASSERT_EFI_ERROR` seems to do nothing and thus can't be used in place of proper error handling.  It's OK for things like `gBS->CloseEvent()`, but return status of `gBS->WaitForEvent` and `gST->ConIn->ReadKeyStroke` define whether they set output parameters to something and you don't want to branch on random values on the stack if they weren't initialized.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+            NULL,
+            &Size,
+            &RequestedActivePcrBanks
+            );
+  // It's zero by default, that means there's been no request to change the PCR banks.
+  if (RequestedActivePcrBanks == 0) {
+    return;
+  } else if (RequestedActivePcrBanks == ActivePcrBanks) {
+    // 
+    // If they're equal, zero it out again and proceed with normal boot.
```
SergiiDmytruk: Why not just delete it (`gRT->SetVariable(REQUESTED_ACTIVE_PCR_BANKS_VARIABLE_NAME, &gEfiTcg2PhysicalPresenceGuid, 0, 0, NULL);`)?  Also I think it can be deleted unconditionally after reading it successfully to not bother deleting it below on <kbd>Escape</kbd>.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+          sizeof(RequestedActivePcrBanks),
+          &RequestedActivePcrBanks
+        );
+    return;
+  }
+  // 
+  // If they're not equal, display the popup and switch to a single bank
+  // of user's choice.
+  // 
+  Status = gBS->CreateEvent(EVT_TIMER, TPL_CALLBACK, NULL, NULL, &TimerEvent);
```
SergiiDmytruk: You're not doing anything when the timer is signaled, so no need to even create one.  The real effect of using it is a blinking UI it seems.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+    UnicodeSPrint(
+      token, sizeof(token),
+      L" %u) %s", (UINT32)(i + 1), mHashAlgs[AvlIdx[i]].Name
+    );
+    StrCatS(OptLine, ARRAY_SIZE(OptLine), token);
+  }
+  
+  while (1) {
+    CHAR16 HintBuf[64];
+    UnicodeSPrint(HintBuf, sizeof(HintBuf), L"Press 1..%u to select a bank.", (UINT32)OptionCount);
```
SergiiDmytruk: These lines aren't needed, the variable isn't referenced below.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+      }
+
+      RequestedActivePcrBanks = SelectedMask;
+      gRT->SetVariable(
+          L"RequestedActivePcrBanks",
+          &gEfiTcg2PhysicalPresenceGuid,
+          EFI_VARIABLE_NON_VOLATILE | EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS,
+          sizeof(RequestedActivePcrBanks),
+          &RequestedActivePcrBanks
+      );
```
SergiiDmytruk: `SelectedMask` is guaranteed to have exactly one bit set here, so in case of a failed operation the dialog about attempt to select multiple banks seems to not really apply.  If the variable is removed above, this paragraph can just be dropped.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
SergiiDmytruk: We can change this if client won't like an extra dialog.  Doesn't seem like a major inconvenience to me either.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,221 @@ WarnIfRecoveryBoot (
...
+        78,
+        9,
+        L"!!! WARNING !!!",
+        L"",
+        L"Multiple PCR banks have been selected, but the current TPM supports",
+        L"only one active bank at a time.",
+        L"",
+        OptLine,
+        L"",
+        L"Press ESC to deny and revert to the default bank.",
```
SergiiDmytruk: This line looks incorrect.  I don't think there is a default bank and we're basically just leaving the state as is.

SergiiDmytruk (changes requested):

SecurityPkg/Library/DxeTcg2PhysicalPresenceLib/DxeTcg2PhysicalPresenceLib.c
filipleple: Nope, no reason I think. [Removed](https://github.com/Dasharo/edk2/pull/282/files#diff-f206868b84cc46d1b4f95b14863dbcd2afd8aed2d6cef61d64d0dcc2991a9f44R310) in all instances

SecurityPkg/Library/DxeTcg2PhysicalPresenceLib/DxeTcg2PhysicalPresenceLib.c
filipleple: Nope, I've missed it. Added in both [SecurityPkg](https://github.com/Dasharo/edk2/pull/282/files#diff-61ad4528c4e36ab981df94d17e9b049393b7c623a6ed2edcc6dbeb4677277234R230) and [OvmfPkg](https://github.com/Dasharo/edk2/pull/282/files#diff-f206868b84cc46d1b4f95b14863dbcd2afd8aed2d6cef61d64d0dcc2991a9f44R337)

SecurityPkg/Include/Guid/Tcg2PhysicalPresenceData.h
filipleple: [Using](https://github.com/Dasharo/edk2/pull/282/files#diff-61ad4528c4e36ab981df94d17e9b049393b7c623a6ed2edcc6dbeb4677277234R232) the macro in all Set&GetVariable calls now

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: Ahh, right, this simplifies a lot. I wasn't sure if I could reliably remove a variable once set, or just change its state. [Switched](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1124) the logic to use EFI_NOT_FOUND.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: Right, [removed](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1142) the timer. UI no longer blinks :D

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: Right, [removed](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1170)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Dropped](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1205)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Fixed](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1181) to explain we're keeping the previous bank state

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: Oh, right. Thanks for the explanation, [fixed](https://github.com/Dasharo/edk2/pull/282/commits/96d8f38078a39815530b8a2d72fceb19248810ab) to retry/return on failure

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,168 @@ WarnIfRecoveryBoot (
...
+  UINTN          i;
+  UINTN          OptionCount = 0;
+  UINTN          AvlIdx[ARRAY_SIZE(mHashAlgs)] = {0};
+  CHAR16         OptLine[81];
+  UINT32         SelectedMask;
+
+  Status = Tpm2GetCapabilitySupportedAndActivePcrs(&TpmHashAlgorithmBitmap, &ActivePcrBanks);
+  if (EFI_ERROR(Status)) {
+      return;
+  }
```
SergiiDmytruk: I think placing these 4 lines after `gRT->SetVariable` is better.  This avoids bothering TPM on boots where PCR banks aren't being changed (so most of the boots) and prevents potential delays associated with that.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,168 @@ WarnIfRecoveryBoot (
...
+            &gEfiTcg2PhysicalPresenceGuid,
+            NULL,
+            &Size,
+            &RequestedActivePcrBanks
+            );
+  // 
+  // If the variable doesn't exist, there's been no request to change the
+  // PCR banks.
+  // 
+  if (Status == EFI_NOT_FOUND) {
```
SergiiDmytruk:
```suggestion
  if (EFI_ERROR(Status)) {
```
since we don't have any data to use in case of other errors as well.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,168 @@ WarnIfRecoveryBoot (
...
+    StrCatS(OptLine, ARRAY_SIZE(OptLine), token);
+  }
+  
+  while (1) {
+    CreateMultiStringPopUp(
+        78,
+        9,
+        L"!!! WARNING !!!",
+        L"",
+        L"Multiple PCR banks have been selected, but the current TPM supports",
```
SergiiDmytruk: `apparently supports` is probably a more accurate wording, since we're inferring this from an unexpected state instead of knowing for sure.

We should probably also check that `Multiple PCR banks have been selected` is in fact true in case mask is different for some other reason.  Adding
```c
if ((RequestedActivePcrBanks & (RequestedActivePcrBanks - 1)) != 0)
    return;
```
near the top should do. This resets the least significant bit set, `PopCount(RequestedActivePcrBanks) > 1` would be more explicit, but there may be no such function in EDK or it has some non-obvious name.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,168 @@ WarnIfRecoveryBoot (
...
+        L"Press ESC to stay with the previously active bank.",
+        L""
+    );
+
+    Status = gBS->WaitForEvent(1, Events, &Index);
+    if (EFI_ERROR(Status)) {
+      continue;
+    }
+    
+    if (Index != 0) continue;
```
SergiiDmytruk: This isn't necessary for a single event, on success this should be zero.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,168 @@ WarnIfRecoveryBoot (
...
+      if (SelectedMask == ActivePcrBanks) {
+        break;
+      }
+
+      Tcg2PhysicalPresenceLibSubmitRequestToPreOSFunction(
+          TCG2_PHYSICAL_PRESENCE_SET_PCR_BANKS,
+          SelectedMask
+      );
+
+      ASSERT_EFI_ERROR(Status);
```
SergiiDmytruk: This should never assert and can be removed.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -1067,6 +1070,168 @@ WarnIfRecoveryBoot (
...
+    Status = gBS->WaitForEvent(1, Events, &Index);
+    if (EFI_ERROR(Status)) {
+      continue;
+    }
+    
+    if (Index != 0) continue;
+
+    Status = gST->ConIn->ReadKeyStroke(gST->ConIn, &Key);
+    if (EFI_ERROR(Status)) {
+      continue;
```
SergiiDmytruk: Continuing on error here and for `WaitForEvent` may not be the best approach.  These functions shouldn't fail randomly and if this is something permanent the code will just hang.  Breaking out of the loop seems more preferable.

SergiiDmytruk (changes requested):
A few small adjustments are necessary.

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Moved](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1133)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Changed](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1122)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Removed](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1194)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Removed](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1216)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Breaking](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1192)

DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
filipleple: [Reworded](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1182) and [added](https://github.com/Dasharo/edk2/pull/282/files#diff-306afacd6b8e1d90f20c52fe64cf2eab3b6d949cdf5225d2b310a06c84a950bcR1142) the multiple bank check. Looks like it's gonna be `== 0` (exit if only a single bit set), not `!= 0`

SergiiDmytruk: Right, the condition was inverted.

SergiiDmytruk (approved):

PR #284, opened 2025-12-01
DasharoPayloadPkg/Library/FmpDeviceSmmLib: Handle read- and write-protected regions

mkopec:
Handle locked regions appropriately by:
- Not failing completely if we could not read some of the firmware
- Skipping writes to locked regions

Note that this assumes that the protected ranges are aligned to the erase/write block size, otherwise the start/end of unprotected ranges might not get updated.

It also depends on the FMAP being consistent with the IFD.

*ref: ncm-2076*

### DIFF
--- a/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
+++ b/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
@@ -80,7 +80,7 @@ GetFmap (
 VOID *
 EFIAPI
 ReadCurrentFirmware (
-  VOID
+  BOOLEAN IsDescriptorLocked
   )
 {
   EFI_STATUS  Status;
@@ -140,8 +140,15 @@ ReadCurrentFirmware (
         Block * BlockSize,
         Status
         ));
-      FreePool (Image);
-      return NULL;
+
+      // We allow for this failure if descriptor is locked because when locked,
+      // certain regions of the flash are read-protected. We can't know which
+      // ones they are until we've read enough of the firmware to get to the
+      // FMAP.
+      if (!IsDescriptorLocked) {
+        FreePool (Image);
+        return NULL;
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -610,6 +610,90 @@ MigrateSmbiosData (
...
+      "%a(): failed to parse current firmware\n",
+      __FUNCTION__
+      ));
+    return FALSE;
+  }
+
+  Region = FmapFindArea(FlashMap, "SI_DESC");
+
+  // There is no descriptor, so everything should be unlocked.
+  if (!Region)
```
mkopec, line 663: If we got here but the DescriptorWriteable var exists, something has gone pretty wrong.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -610,6 +610,90 @@ MigrateSmbiosData (
...
+    return TRUE;
+
+  // Range overlaps locked descriptor.
+  if (RangeOffset + RangeLen > Region->offset &&
+      Region->offset + Region->size > RangeOffset)
+    return FALSE;
+
+  Region = FmapFindArea(FlashMap, "SI_ME");
+
+  // There is no ME (?)
```
mkopec: Are there even platforms with a descriptor but without ME? Maybe embedded platforms with TXE?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1009,6 +1009,17 @@ FmpDeviceSetImageWithStatus (
       continue;
     }
 
+    // Only touch blocks that are fully unlocked for writing.
+    if (!IsRangeWriteable(CurrentImage, BlockSize * BlockCount, Offset, BlockSize)) {
```
SergiiDmytruk:
```suggestion
    if (!IsRangeWriteable(CurrentImage, ImageSize, Offset, BlockSize)) {
```

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1009,6 +1009,17 @@ FmpDeviceSetImageWithStatus (
       continue;
     }
 
+    // Only touch blocks that are fully unlocked for writing.
+    if (!IsRangeWriteable(CurrentImage, BlockSize * BlockCount, Offset, BlockSize)) {
+      DEBUG ((
+        DEBUG_ERROR,
```
SergiiDmytruk:
```suggestion
        DEBUG_INFO,
```
Since we expect this to happen and keep going assuming the update will succeed, I think this should be `INFO`.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -610,6 +611,93 @@ MigrateSmbiosData (
...
+      &DescriptorWriteable
+      );
+
+  // Variable does not exist on platforms without a descriptor (e.g. AMD)
+  if (EFI_ERROR(Status))
+    DescriptorWriteable = TRUE;
+
+  // Everything is unlocked.
+  if (DescriptorWriteable)
+    return TRUE;
```
SergiiDmytruk: Reading variables from `SMMSTORE` (even though in practice variables will get cached) in the middle of overwriting it doesn't seem safe.  We also don't need to do it every time.  So I would move determining whether anything should be unreadable/unwritable to `FmpDeviceSetImageWithStatus()` (or rather a function near it) and then:
 * skipped calling `IsRangeWriteable()` if the descriptor (and by assumption everything else) is writable
 * passed a flag to `ReadCurrentFirmware()` to skip read failures conditionally

This way we'll also preserve current behaviour for non-fused platforms which might be helpful to catch some errors.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1009,6 +1009,17 @@ FmpDeviceSetImageWithStatus (
       continue;
     }
```
SergiiDmytruk, line 1040: To not leave this information only on GitHub:

```suggestion
    // This assumes that the protected ranges are aligned to SMMSTORE's block size,
    // otherwise the start/end of unprotected ranges might not get updated.
```

Note that this is not granularity of a flash chip, the value is [hard-coded](https://github.com/Dasharo/coreboot/blob/43c572154ea74bbe24b6d67f0ebbb9498499e0e7/src/include/smmstore.h#L41-L47) to be 64 KiB in coreboot.  Is this a problem for `SI_GBE` then?
```
	SI_ALL 10M {
		SI_DESC 16K
#if CONFIG_MAINBOARD_USES_IFD_GBE_REGION
		SI_GBE 8K
#endif
		SI_ME
	}

```

SergiiDmytruk (dismissed):

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: Hmm thankfully we don't need to touch this region. Currently we're migrating it but not touching it should work too. I'll add it to the list in IsRangeWriteable

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/commit/31b804f996ad851a20838d3d7d4179840466ea01

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/commit/31b804f996ad851a20838d3d7d4179840466ea01

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/commit/31b804f996ad851a20838d3d7d4179840466ea01

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/commit/31b804f996ad851a20838d3d7d4179840466ea01

filipleple: LGTM, Descriptor lock state is read once via IsDescriptorLocked(); if locked, ReadCurrentFirmware tolerates read errors and block writes are filtered using FMAP-based IsRangeWriteable() instead of per-block SMMSTORE variable reads.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: LGTM, SI_GBE is in the exclusion list in IsRangeWriteable and we're skipping any overlapping blocks when DescriptorLocked is true.

filipleple (approved):

PR #285, opened 2025-12-09
Check BtG key before updating via Capsule Update

mkopec:
Rejects updating via Capsule Update if BtG is enabled and the updated image does not have the same OEM Root Key in the key manifest.

*ref: ncm-2095*


DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+
+  @param[in] Image     Image to check
+  @param[in] ImageLen  Size of the image to check
+
+  @return TRUE    If Boot Guard BPM and KM are present
+  @return FALSE   If Boot Guard BPM and KM are absent
+**/
+STATIC
+BOOLEAN
+IsBootGuardEnabled(
```
miczyg1:
```suggestion
IsBootGuardEnabled (
```

Please keep the project coding style.

There are lots of places where function calls miss the space before `(`

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+      ));
+    return FALSE;
+  }
+
+  if (!GetCbfs (Image, FlashMap, &Cbfs)) {
+    DEBUG ((DEBUG_ERROR, "%a(): failed to load CBFS\n", __FUNCTION__));
+    return FALSE;
+  }
+
+  File = cbfs_get_entry (&Cbfs, "key_manifest.bin");
```
miczyg1, line 932: These filenames are configurable in coreboot Kconfig. Would be wise to pass them to EDK2 build system as PCDs, and then get them with `GetPcdPtr`.

Otherwise, we may find out one day, by surprise, that this doesn't work, becasue CBFS filenames are different somewhere else.

Also I thought we talked about empty placeholder manifests. They may be present in the image, which doesn;t necessarily mean a provisioned image. What if placeholder manifests are present? You should at least check if these manifests have a proper signature `__KEYM__` or `__ACBP__`.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  if (File == NULL)
+    return FALSE;
+
+  File = cbfs_get_entry (&Cbfs, "boot_policy_manifest.bin");
```
miczyg1, line 936: a/a

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  IN CONST UINTN  KmSize,
+  OUT VOID        **OemRootKey,
+  OUT UINTN       *OemRootKeySize
+  )
+{
+  BtgPubKey *PubKeyOffset, *PubKey;
+  KeyManifestHeader *Km = (KeyManifestHeader*) KmBuffer;
+
+  if (Km == NULL ||
+      OemRootKeySize == NULL ||
```
miczyg1: Add check if OemRootKey is NULL, it is both input and output parameter

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  @return EFI_INVALID_PARAMETER    If the input parameters or values in the Key
+                                   Manifest are invalid
+  @return EFI_SUCCESS              If extraction was successful
+**/
+STATIC
+EFI_STATUS
+GetOemRootKeyFromKm (
+  IN CONST VOID   *KmBuffer,
+  IN CONST UINTN  KmSize,
+  OUT VOID        **OemRootKey,
```
miczyg1:
```suggestion
  IN OUT VOID        **OemRootKey,
```

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+    GroupName = GetSmbiosString (TempStruct, Type14->GroupName);
+    if (GroupName != NULL && AsciiStrCmp (GroupName, "$MEI") == 0) {
+      TargetHandle = Type14->ItemHandle;
+      FoundTargetHandle = TRUE;
+      break;
+    }
+  }
+
+  if (!FoundTargetHandle)
+    return FALSE;
```
miczyg1: Is it safe to assume unfused platform if we can't get info about fuse state?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+     FoundTargetHandle = FALSE;
+     do {
+       Status = Smbios->GetNext (Smbios, &SmbiosHandle, NULL, &SmbiosRecord, NULL);
+       if (!EFI_ERROR(Status) && SmbiosRecord->Handle == TargetHandle) {
+         FoundTargetHandle = TRUE;
+         break;
+       }
+     } while (!EFI_ERROR(Status));
+
+     if (!FoundTargetHandle) return FALSE;
```
miczyg1: a/a

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  }
+
+  RawData = (UINT8 *)SmbiosRecord;
+
+  if (SmbiosRecord->Length < (sizeof(UINT32)))
+    return FALSE;
```
miczyg1: a/a

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+
+  CopyMem (&Hfsts1, &RawData[0], sizeof(UINT32));
+
+  if ((Hfsts1 & BIT4) == 0)
```
miczyg1: BIT4 is the SPI descriptor lock(ME > 11)/EOM bit (ME11) state, not the fuse state right?

We should check the bit 30 in hfsts6

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.h
```diff
@@ -46,4 +46,24 @@ IsRangeWriteable (
...
+**/
+BOOLEAN
+EFIAPI
+AreImageBtgKeysCompatible (
+  IN CONST VOID  *Current,
+  IN CONST VOID  *Updated,
+  IN CONST UINTN ImageSize
+  );
+
+#endif // FLASHING_H__
```
miczyg1: Fix new line please

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1015,6 +1016,17 @@ FmpDeviceSetImageWithStatus (
 
   IncrementProgress (Progress, TotalSteps, ReadSteps, &Step, &ShouldReportProgress);
 
+  if (!AreImageBtgKeysCompatible(CurrentImage, Image, ImageSize)) {
+    FreePool (CurrentImage);
+    AsciiPrint("New image is not signed with a compatible OEM Root Key, refusing to flash\n");
```
miczyg1:
```suggestion
    AsciiPrint("New image is not signed with a compatible OEM Root Key, aborting update\n");

```

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  )
+{
+  EFI_STATUS         Status;
+  struct cbfs_file   *CurrentKmFile, *UpdatedKmFile;
+  struct cbfs_image  CurrentCbfs, UpdatedCbfs;
+  CONST Fmap         *CurrentFlashMap, *UpdatedFlashMap;
+  VOID               *CurrentOemKey, *UpdatedOemKey;
+  UINTN              CurrentOemKeyLen, UpdatedOemKeyLen;
+
+  // Platform is unfused, so different BtG key doesn't matter.
```
miczyg1: That's only true if ME region is not locked. Shouldn't we check if ME is unlocked to flash an image with different keys?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1015,6 +1016,17 @@ FmpDeviceSetImageWithStatus (
 
   IncrementProgress (Progress, TotalSteps, ReadSteps, &Step, &ShouldReportProgress);
 
+  if (!AreImageBtgKeysCompatible(CurrentImage, Image, ImageSize)) {
+    FreePool (CurrentImage);
+    AsciiPrint("New image is not signed with a compatible OEM Root Key, refusing to flash\n");
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a(): New image is not signed with a compatible OEM Root Key, refusing to flash\n",
```
miczyg1:
```suggestion
      "%a(): New image is not signed with a compatible OEM Root Key, aborting update\n",

```

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
miczyg1: Sounds less agressive I think

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  OUT UINTN       *OemRootKeySize
+  )
+{
+  BtgPubKey *PubKeyOffset, *PubKey;
+  KeyManifestHeader *Km = (KeyManifestHeader*) KmBuffer;
+
+  if (Km == NULL ||
+      OemRootKeySize == NULL ||
+      KmSize < sizeof(KeyManifestHeader) ||
+      KmSize < Km->KeySignatureOffset + sizeof(KeyAndSigHeader) + sizeof(BtgPubKey))
```
miczyg1: Maybe add a check for KM signature: `__KEYM__` if the manifest is even valid. Because KeySignatureOffset may be 0, so this condition may be satisfied when a placeholder KM is used.

miczyg1 (dismissed):

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +835,198 @@ MergeFirmwareImages (
...
+  if (PubKeyOffset->KeySize != 2048 && PubKeyOffset->KeySize != 3072)
+    return EFI_INVALID_PARAMETER;
+
+  if (PubKeyOffset->Exponent != 0x10001)
+    return EFI_INVALID_PARAMETER;
+
+  *OemRootKeySize = sizeof(BtgPubKey) + PubKeyOffset->KeySize / 8;
+
+  PubKey = AllocatePool (*OemRootKeySize);
+  CopyMem(PubKey, PubKeyOffset, *OemRootKeySize);
```
SergiiDmytruk: There is a convenient shortcut for making a copy: `PubKey = AllocateCopyPool(PubKeyOffset, *OemRootKeySize);`.  However, why use dynamic allocation at all?  You're only reading the data inside of images while they exist.  As is, the code leaks the memory allocated here.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +835,198 @@ MergeFirmwareImages (
...
+      KmSize < Km->KeySignatureOffset + sizeof(KeyAndSigHeader) + sizeof(BtgPubKey))
+    return EFI_INVALID_PARAMETER;
+
+  PubKeyOffset = (BtgPubKey*)((UINT8*)Km + Km->KeySignatureOffset + sizeof(KeyAndSigHeader));
+
+  if (PubKeyOffset->KeySize != 2048 && PubKeyOffset->KeySize != 3072)
+    return EFI_INVALID_PARAMETER;
+
+  if (PubKeyOffset->Exponent != 0x10001)
+    return EFI_INVALID_PARAMETER;
```
SergiiDmytruk: Are these checks done for validation that this is in fact a key?  Checking fields of `KeyAndSigHeader` should avoid processing keys of unhandled types (e.g., ECC) and not limit key size or demand a specific exponent (even though it will probably be `0x10001`).

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -930,6 +933,126 @@ GetOemRootKeyFromKm (
...
+  SMBIOS_TABLE_TYPE14_SINGLE_ITEM *Type14;
+  CHAR8                      *GroupName;
+  UINT16                     TargetHandle;
+  BOOLEAN                    FoundTargetHandle;
+  UINT8                      *RawData;
+  UINT32                     Hfsts1;
+
+  Status = gBS->LocateProtocol (&gEfiSmbiosProtocolGuid, NULL, (VOID **)&Smbios);
+  if (EFI_ERROR (Status))
+    return FALSE;
```
SergiiDmytruk: Unfused if no SMBIOS protocol is found?  This shouldn't even happen, but sounds somewhat weird.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+  FoundTargetHandle = FALSE;
+
+  while (!EFI_ERROR (Smbios->GetNext (Smbios, &SmbiosHandle, &SmbiosType, &SmbiosRecord, NULL))) {
+    Type14 = (SMBIOS_TABLE_TYPE14_SINGLE_ITEM *) SmbiosRecord;
+
+    SMBIOS_STRUCTURE_POINTER TempStruct;
+    TempStruct.Hdr = SmbiosRecord;
+
+    GroupName = GetSmbiosString (TempStruct, Type14->GroupName);
+    if (GroupName != NULL && AsciiStrCmp (GroupName, "$MEI") == 0) {
```
SergiiDmytruk, line 1065: Do we want to check that `Type14->ItemType` is `0xdb` (arbitrary value set in coreboot?)?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +838,320 @@ MergeFirmwareImages (
...
+
+     if (!FoundTargetHandle) return FALSE;
+  }
+
+  RawData = (UINT8 *)SmbiosRecord;
+
+  if (SmbiosRecord->Length < (sizeof(UINT32)))
+    return FALSE;
+
+  CopyMem (&Hfsts1, &RawData[0], sizeof(UINT32));
```
SergiiDmytruk: This seems wrong.  `RawData` points at the start of the record, not its data.  Also, the whole record looks like this:
```c
struct fwsts_record {
	u8 heci_name;
	u32 reg[6];
} __packed;

struct fwsts_smbios_table {
	struct smbios_header header;
	u8 version;
	u8 count;
	struct fwsts_record record[CONFIG_MAX_MEI_DEVICES];
	u8 eos[2];
} __packed;
```
So need to reach into `record[0].reg[5]` (HFSTS6; always first HECI?).  Currently the header seems to be copied.

SergiiDmytruk (changes requested):
Came back after looking at coreboot changes.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: fb3363242da4043e98652c308c0ab7727de41c80

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.h
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/0339bc0879b7d10e025d81012f0123c045826f6b

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: ad8604f8c0ae5a8a7dcdd794ed2d15e2d3971295

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: d35d3a6c88263639d9de327cda2f3f9b6cd28f3f

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: d35d3a6c88263639d9de327cda2f3f9b6cd28f3f

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/e7b8bf9bc82c192f2e49b8ce5df3d3f04d26e336

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/e7b8bf9bc82c192f2e49b8ce5df3d3f04d26e336

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/5e0cbbb5485e7bd6ffc69d8b88336188423315ec

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/88b3336bdd555a31fa6fcd7ee633295c4c2bca65

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: Not sure. @miczyg1 mentioned this could be different on other BIOS implementations but that doesn't matter much here.

miczyg1: > Do we want to check that Type14->ItemType is 0xdb (arbitrary value set in coreboot?)?

We hardcode that to 0xdb, but the type is BIOS-build-dependent, when it comes to proprietary BIOS. Also structrues of given type may occur more than once. `Handle` is what uniquely identifies the entry we search for.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/3b5485a046b524fefcda29b2f758bd66fe44a493

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/0908ead1d13b3c535ec986d06ad10f98d461924e for the KM signature check

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/0908ead1d13b3c535ec986d06ad10f98d461924e

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: Yes these checks are for validation if this is indeed a key. Changed to checking the key type instead (RSA): https://github.com/Dasharo/edk2/pull/285/commits/be5faada8f25f1ff84a9b9d3cda0c5c6483dac15

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: The filenames aren't actually configurable in coreboot's buildsystem

miczyg1: Ahh sorry. That was probably the ACM binary names that were configurable...

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: Changed to TRUE, let's be careful and assume the platform is fused if we could not determine otherwise: https://github.com/Dasharo/edk2/pull/285/commits/10fe09c0b797c160e8f8e4e874b84c97ba39be83

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: Changed to TRUE, let's be careful and assume the platform is fused if we could not determine otherwise: https://github.com/Dasharo/edk2/pull/285/commits/10fe09c0b797c160e8f8e4e874b84c97ba39be83

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: Changed to TRUE, let's be careful and assume the platform is fused if we could not determine otherwise: https://github.com/Dasharo/edk2/pull/285/commits/10fe09c0b797c160e8f8e4e874b84c97ba39be83

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: Changed to TRUE, let's be careful and assume the platform is fused if we could not determine otherwise: https://github.com/Dasharo/edk2/pull/285/commits/10fe09c0b797c160e8f8e4e874b84c97ba39be83

SergiiDmytruk (dismissed):
Needs testing, but looks good now.  coreboot PR will need to have EDK hash updated.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -992,12 +1371,20 @@ FmpDeviceSetImageWithStatus (
...
       __FUNCTION__
       ));
+    ErrorString = L"Intel Boot Guard OEM Root Key mismatch";
+    if (AbortReason != NULL) {
+        *AbortReason = AllocateCopyPool(StrSize(ErrorString), ErrorString);
+    }
+
+    ShowBtgErrorPopup(CurrentImage, ImageSize, ErrorString);
+
+    FreePool(ErrorString);
```
SergiiDmytruk: No need to free a string literal.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,6 +865,380 @@ QueryVariableInfoHook (
...
+        L"Abort reason:",
+        ErrorMessage,
+        L"",
+        L"Fused OEM RK Hash:",
+        CurrentOemRkString,
+        L"",
+        L"Press ENTER to boot normally."
+        );
+
+    Status = gBS->WaitForEvent (1, Events, &Index);
```
SergiiDmytruk, line 1242: `Events` should also include `TimerEvent`.  As is, this `while` loop will likely never stop executing.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -9,10 +9,12 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 **/
 
-#include <PiDxe.h>
+#include <Coreboot.h>
 #include <Guid/SystemResourceTable.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/BlParseLib.h>
+#include <Library/BootLogoLib.h>
```
SergiiDmytruk: I think this library is not actually used but gets added here and in the `.inf`-file.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,6 +865,380 @@ QueryVariableInfoHook (
...
+    //
+    // just throw away Key
+    //
+  }
+}
+
+// Configuration
+#define SCALE_FACTOR        2
+#define GLYPH_WIDTH         8
+#define GLYPH_HEIGHT        19
```
SergiiDmytruk, line 887: Glyph dimensions are fixed by the font?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,6 +865,380 @@ QueryVariableInfoHook (
...
+  //
+  while (!EFI_ERROR (gST->ConIn->ReadKeyStroke (gST->ConIn, &Key))) {
+    //
+    // just throw away Key
+    //
+  }
+}
+
+// Configuration
+#define SCALE_FACTOR        2
```
SergiiDmytruk: Should this be determined by looking at resolution?  If it's HiDPI, then scale by a factor of 2 and don't otherwise.  I'm just unsure we'll notice an enormous dialog during testing for some other devices unless we'll test this case.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,6 +865,380 @@ QueryVariableInfoHook (
...
+  EFI_FONT_DISPLAY_INFO FontInfo;
+  ZeroMem(&FontInfo, sizeof(EFI_FONT_DISPLAY_INFO));
+  FontInfo.ForegroundColor = (EFI_GRAPHICS_OUTPUT_BLT_PIXEL)POPUP_TEXT_COLOR;
+  FontInfo.BackgroundColor = (EFI_GRAPHICS_OUTPUT_BLT_PIXEL)POPUP_BG_COLOR;
+
+  // 6. Render Text Loop
+  UINTN CurrentY = BoxY + (PADDING_Y * SCALE_FACTOR);
+  UINTN TextStartX = BoxX + (PADDING_X * SCALE_FACTOR);
+
+  VA_START(Args, NumberOfLines);
```
SergiiDmytruk: Can't do this twice (it may work, but it's not guaranteed).  Should use `VA_COPY` after the first `VA_START` and before `VA_END`.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -1099,6 +1100,100 @@ IsPlatformFused (
...
+  );
+
+  if (EFI_ERROR (Status))
+    return Status;
+
+  HashInputLen = (OemKey->KeySize / 8) + sizeof(OemKey->Exponent);
+  HashInput = AllocatePool (HashInputLen);
+
+  CopyMem (HashInput, OemKey->Modulus, OemKey->KeySize / 8);
+  CopyMem (HashInput + (OemKey->KeySize / 8), &OemKey->Exponent, sizeof(OemKey->Exponent));
```
SergiiDmytruk, line 1180: Were modulus and exponent swapped for some cryptographic reason?

SergiiDmytruk (changes requested):
A few things to fix.  Haven't tested the dialog yet, will also need to go through its implementation in more detail.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: That's just how Intel does it, I don't know what the justification for this is. The hash calculated here matches what's shown in Intel MFIT in the OEM Key Hash field.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/9bb64f060082cd5cc5ee86e248213242320127a6

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/170d6b39ccff98e61582de6a7547bc012bdce27a

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
SergiiDmytruk: Then a comment saying that ordering of fields and digest algorithm matches behaviour of Intel's MFIT could be helpful.  The fact of hashing looked weird as well, but I thought it's done to shorten warning message.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/8c7a3ef7f3688feb5d410ae92573b36403c15146

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: It should actually never stop unless enter is explicitly pressed. We want the user to confirm they've read the message. Will clean up

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: It's fixed in the font and also defined like this in a couple of places in the edk2 tree

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/4a1c766b7e6f312305409a6717ea25d003ebd6bd

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: Made scale factor adjustable: https://github.com/Dasharo/edk2/pull/285/commits/4b07561bb88efe003204a25551871e14217ea1a3

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
SergiiDmytruk: The code still creates a timer event.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: and made it so that if HRes>1920 only then do we scale 2x: https://github.com/Dasharo/edk2/pull/285/commits/fae3a925a4aaa3134049a6dc1c70b8d98b74e05d

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: whoops sorry: https://github.com/Dasharo/edk2/pull/285/commits/8d3ee0305e4dde96d4a4656479f3c33476924e19

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/edf5f1a73ab14232c7e38ec30857bffe6b16d586

The key is basically hashed to reduce the number of E-fuses needed for Boot Guard, as I understand it. The hash is what is actually burnt into FPFs.

mkopec:
Looks like this: 
image

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,29 +865,392 @@ QueryVariableInfoHook (
...
+
+  DrainInput ();
+
+  Events[0] = gST->ConIn->WaitForKey;
+
+  do {
+    ClearScreen();
+    CreateMultiStringPopUpScaled (
+        GetDisplayScaleFactor(),
+        100,
```
SergiiDmytruk:
```suggestion
        96,
```
With `100`:
```
ScreenWidth=800
ScreenHeight=600
RequestedWidth=100
UnscaledContentW=800
UnscaledContentH=247
BoxW=832
BoxH=285
```

Image

While `96` fits perfectly:
```
RequestedWidth=96
ScreenWidth=800
ScreenHeight=600
RequestedWidth=96
UnscaledContentW=768
UnscaledContentH=247
BoxW=800
BoxH=285
```

Image

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,29 +865,392 @@ QueryVariableInfoHook (
...
+  Gop->Blt(Gop, &BorderColor, EfiBltVideoFill, 0, 0, BoxX, BoxY, BoxW, BORDER_WIDTH, 0); // Top
+  Gop->Blt(Gop, &BorderColor, EfiBltVideoFill, 0, 0, BoxX, BoxY + BoxH - BORDER_WIDTH, BoxW, BORDER_WIDTH, 0); // Bottom
+  Gop->Blt(Gop, &BorderColor, EfiBltVideoFill, 0, 0, BoxX, BoxY, BORDER_WIDTH, BoxH, 0); // Left
+  Gop->Blt(Gop, &BorderColor, EfiBltVideoFill, 0, 0, BoxX + BoxW - BORDER_WIDTH, BoxY, BORDER_WIDTH, BoxH, 0); // Right
+
+  // 5. Prepare for Text Rendering
+  // We allocate enough space for one line at a time
+  UINTN LinePixels1x = UnscaledContentW * GLYPH_HEIGHT;
+  RenderBuffer1x = AllocateZeroPool(LinePixels1x * sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
+  RenderBufferScaled = AllocateZeroPool(LinePixels1x * ScaleFactor * ScaleFactor * sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
```
SergiiDmytruk, line 1074: Something seems to be trashing memory:
```
DXE_ASSERT!: [DxeCore] .../coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Core/Dxe/Mem/Pool.c (741): Tail->Signature == ((('p') | ('t' << 8)) | ((('a') | ('l' << 8)) << 16))

DXE_ASSERT!: [DxeCore] .../coreboot/payloads/external/edk2/workspace/Dasharo/MdeModulePkg/Core/Dxe/Mem/Pool.c (742): Head->Size == Tail->Size

ASSERT_EFI_ERROR (Status = Invalid Parameter)

DXE_ASSERT!: [FmpDxe] .../coreboot/payloads/external/edk2/workspace/Dasharo/MdePkg/Library/UefiMemoryAllocationLib/MemoryAllocationLib.c (821): !(((RETURN_STATUS)(Status)) >= 0x8000000000000000ULL)
```

If I allocate 10 times the amount on this line the errors are gone, but I don't see this buffer being overrun in the code.  Maybe the issue in a different place.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: https://github.com/Dasharo/edk2/pull/285/commits/b3f0caeafcb3cde4d3bf8d39d1cfae12a7c9fb10

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
mkopec: I'm not sure what to do here, how to address this, @SergiiDmytruk do you have suggestions?

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -861,29 +865,392 @@ QueryVariableInfoHook (
...
+    for (x = 0; x < SourceWidth; x++) {
+      EFI_GRAPHICS_OUTPUT_BLT_PIXEL Pixel = Source[y * SourceWidth + x];
+
+      // Calculate index of the top-left pixel in the scaled destination block
+      UINTN DestIndex = (y * ScaleFactor) * DestWidth + (x * ScaleFactor);
+
+      // Fill the scaled block
+      for (UINTN Index = 0; Index < ScaleFactor; ++Index) {
+        Dest[DestIndex + Index]             = Pixel;
+        Dest[DestIndex + DestWidth + Index] = Pixel;
```
SergiiDmytruk: Here is the cause of memory error.  This assumes scale factor of 2.  Need to add another loop over `ScaleFactor` to update appropriate number of lines in the destination.

mkopec: https://github.com/Dasharo/edk2/pull/285/commits/de7e555a0786ee4ebac0b92d867fca8f69c9696d

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -797,3 +860,394 @@ MergeFirmwareImages (
...
+
+  // Platform is unfused and unlocked, so different BtG key doesn't matter.
+  if (!IsPlatformFused () && !IsDescriptorLocked())
+    return TRUE;
+
+  // Boot Guard is not currently deployed.
+  if (!IsBootGuardEnabled (Current, ImageSize))
+    return TRUE;
+
+  // Going from BtG -> No BtG is not possible if the platform is fused.
```
SergiiDmytruk, line 1232: The comment is for an older version of the code which checked for fused state but not descriptor's lock state.  Now we can get here when:
- platform is fused, descriptor is locked (keys must match)
- platform is fused, descriptor is unlocked (won't happen with current configuration)
- platform is unfused, descriptor is locked (do we need to check BootGuard keys in this case?  is this even possible/relevant?)

SergiiDmytruk (approved):
LGTM, just one question to be sure we've thought about all cases here.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
mkopec: It is relevant if the descriptor is locked, because if the ME is provisioned for some bootguard keys, BtG won't accept an unsigned BIOS region or one signed with other keys. BtG will prevent the platform from booting even though the fuses aren't blown.

If the descriptor was unlocked, we'd just update the ME to match the BIOS and everything would boot fine.

PR #286, opened 2025-12-19
FmpDeviceSmmLib/Flashing.c: write-protect BOOTBLOCK and COREBOOT

filipleple:


DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -475,6 +412,46 @@ MigrateVariables (
...
+{
+DEBUG ((
+  DEBUG_ERROR,
+  "%a(): Entered, will try to merge BOOTBLOCK\n",
+  __FUNCTION__
+  ));
+
+  RegionMigrationStatus  Status;
+
+  Status = MigrateRegion ("BOOTBLOCK", Data, TRUE);
```
mkopec, line 559: Consider that when Top Swap is active, the contents of the BOOTBLOCK and TOPSWAP regions are swapped, but the names in the FMAP still refer to their original positions. So this code has to be aware of whether TS is active and pick the correct region.

mkopec: Oh and the update image is not swapped but the dumped firmware *is*, so you would have to take that into account when migrating the region

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -475,6 +412,46 @@ MigrateVariables (
...
+  )
+{
+  DEBUG ((
+    DEBUG_ERROR,
+    "%a(): Entered, will try to merge COREBOOT\n",
+    __FUNCTION__
+    ));
+  RegionMigrationStatus  Status;
+
+  Status = MigrateRegion ("COREBOOT", Data, TRUE);
```
SergiiDmytruk, line 580: Just to make sure this is by design. Do we intend to only ever update one set of regions instead of updating the inactive one and booting into it?

filipleple: Hmm, yeah IIUC the main value proposition is that resetting CMOS would always revert to the "safe" partition from the "experimental" one, without having to reflash. So the R/O and R/W regions have to be permanently fixed in place

SergiiDmytruk: OK, additionally, seems like in case of error Boot Guard *enables* top-swap rather than toggling or disabling it, so our "golden copy" seems to be `COREBOOT_TS` and `TOPSWAP`. 

In this case these conditions below need  to be removed, because `COREBOOT` can't be absent:
```c
      || Status == REGION_NOT_IN_SRC
      || Status == REGION_NOT_IN_DST;
```

filipleple: Right, removed: d19f9841b781f921d19413634bd7521d17e4144d

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
filipleple: Now it's GetCbfs and GetFmap are Top Swap - aware: https://github.com/Dasharo/edk2/pull/286/changes#diff-df48c35c52e8b3221b927ccc9e290725c37320ea2077f5b0862609923f3a8aa0R161

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
SergiiDmytruk: `BOOTBLOCK` can be absent though.  Either way, we apparently should be updating only `COREBOOT` and `BOOTBLOCK` so this could integrate with Boot Guard and right now:
 - we update `COREBOOT_TS` and `TOPSWAP`
 - migrate data to different regions depending on the state of TS
 - do not set TS option anywhere after performing an update?

Need to sort out what the actual final update flow will look like taking Boot Guard into account and settle on:
 - which regions are updated
 - when TS is modified an how (e.g., only set to `0` if we only update `COREBOOT` and `BOOTBLOCK`)

mkopec: > Either way, we apparently should be updating only COREBOOT and BOOTBLOCK so this could integrate with Boot Guard and right now:

I'm not sure about that. The goal of the feature was to enable going back to a working firmware copy by resetting the CMOS battery. This would be the BOOTBLOCK and COREBOOT regions.

Unfortunately the Top Swap implementation in boot guard doesn't seem to make a lot of sense in this context. It has the automatic switching on verification failure but it should've been toggle-based and not a hardcoded enable of the top swap bit. 

Also i'm not sure if the top swap implementation in Boot Guard works if the BOOTBLOCK region is entirely corrupted  (no readable BPM). I suppose it would need to be tested.

SergiiDmytruk: OK, then I'll assume non-BtG use case taking precedence and I guess BtG can have some workaround later on (or we'll just live with its behaviour).

SergiiDmytruk: Fixed `MigrateBOOTBLOCK` and made various stylistic adjustments while going through the changes in more detail: https://github.com/Dasharo/edk2/compare/53e8bb6731cf8f84628287077b713582ab1eac3f..699f0d638374740b4595c5a2a994d2ccbc4c23cf#diff-df48c35c52e8b3221b927ccc9e290725c37320ea2077f5b0862609923f3a8aa0R561-R563

SergiiDmytruk:
@mkopec, when redundancy is in use, do we need to enforce that FMAP is the same in old and new images?  Right now we're overwriting FMAP during an update, although there are some checks for locations and sizes of regions.  Checking that all FMAP bytes are identical is easy, but maybe we only care about `BIOS` IFD region and thus requiring 100% match is too restricting.  This is capsule's code, so we don't have to do it now, but seems worth mentioning and thinking about it.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
filipleple: OK great, thanks! So I guess I'm resolving this thread after looking through the new changes

SergiiDmytruk:
Another thing: so who's responsibility is it to update slot's flag in CMOS?

filipleple:
@SergiiDmytruk hmm good thought, now that I look at the diagram in https://github.com/Dasharo/dasharo-issues/issues/1468, it looks like it's the platform/firmware responsibility. So I think with capsule update it's possible to add setting the option where coreboot probes whether a capsule is present. In the case of a flashrom update I'm not sure if it'd be so easy to parse both slots and decide if the second one has been updated

SergiiDmytruk:
> So I think with capsule update it's possible to add setting the option where coreboot probes whether a capsule is present.

Not ideal since EDK can reject capsule processing for various reasons, but can work for a start.  Looking at `nvramtool`, setting the flag from EDK would take finding `cmos.layout` in CBFS and parsing it to find where the flag and checksum are in CMOS, then updating both, which doesn't seem that complicated.

> In the case of a flashrom update I'm not sure if it'd be so easy to parse both slots and decide if the second one has been updated

Maybe `nvramtool` is meant to be used in this case after running `flashrom`.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -754,6 +880,30 @@ IsRangeWriteable (
...
     return FALSE;
 
+  // We're only adding a TOPSWAP region if we're using TOP_SWAP_REDUNDANCY.
+  // I think it can be reliably used to gate locking the Slot A regions with
+  // less code than using a new PCD, but also with less flexibility
+  Region = FmapFindArea(FlashMap, "TOPSWAP");
+  if (Region) {
+    // The regions BOOTBLOCK and COREBOOT are to remain read-only golden copies
+    // of the firmware if we're using TOP_SWAP_REDUNDANCY
+    Region = FmapFindArea(FlashMap, "BOOTBLOCK");
```
SergiiDmytruk, line 890: I think we need to account for TS being active here.  We do that when handling image of the flash read via MMIO, but not when we write the new image.  When TS is active, `BOOTBLOCK` is `TOPSWAP` and we want to overwrite it but not touch `TOPSWAP`.

SergiiDmytruk: Actually, no.  MMIO window is read-only so writes are done via SPI.  I thought I somehow saw old firmware after doing a capsule update.

SergiiDmytruk (approved):
Switching between slots will be addressed separately.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -854,6 +1006,16 @@ MergeFirmwareImages (
     goto Fail;
   }
 
+  if (!MigrateBOOTBLOCK (&Data)) {
+    DEBUG ((DEBUG_ERROR, "%a(): MigrateBOOTBLOCK () failed\n", __FUNCTION__));
+    goto Fail;
+  }
+
+  if (!MigrateCOREBOOT (&Data)) {
```
mkopec, line 1014: This broke updating all platforms without redundancy enabled. This needs to be done only if redundancy is actually implemented.

mkopec: https://github.com/Dasharo/edk2/pull/302

PR #287, opened 2026-01-07
Support loading and processing on-disk capsules during the same boot

SergiiDmytruk:
This makes it possible to disable ME via HMRFPO before applying an update, as on-disk capsules don't rely on preserving RAM's state across reboots.

As usual, better review commits in order.

Also fixes use of an uninitialized variable (gets used on every boot) and some memory leaks (although we don't use that code).

First half of https://github.com/Dasharo/dasharo-issues/issues/1438#issuecomment-3703953941 could be used for context.

coreboot PR: https://github.com/Dasharo/coreboot/pull/823
issue: https://github.com/Dasharo/dasharo-issues/issues/1438
*ref: dsh-1130*


macpijan:
Just a high-level question from my side right now: it is entirely configurable and merging this would not impact the behavior of capsule update on existing systems until we opt-in to do so?

SergiiDmytruk:
> Just a high-level question from my side right now: it is entirely configurable and merging this would not impact the behavior of capsule update on existing systems until we opt-in to do so?

It shouldn't.  There are checks via `CoDCheckCapsuleOnDiskFlag()` which now checks for `PcdCapsuleOnDiskSupport` inside and it's `false` by default.  This is something reviewers can have on their mind while looking at the changes.

I've just updated coreboot part to ignore a new variable if it gets set by a user (https://github.com/Dasharo/coreboot/compare/9bf9e7bd450233be6437971e34f10f8dec985a6a..78ce46a6fc786587f2bd52887481520dd0c1e847) to disable some code paths when this isn't enabled.

mkopec:
@SergiiDmytruk how did you test this? Did you manually place the capsule where needed and set the EFI variable by hand, or did you use some sort of existing tool? fwupd seems to have some CoD support: https://github.com/fwupd/fwupd/tree/main/plugins/uefi-capsule#capsule-update-on-disk

mkopec:
Well, fwupd does detect CoD support:

```
├─System Firmware:
│ │   Device ID:          e1dc05aae06da6ff707faf3329b566cfaadab25b
│ │   Summary:            UEFI System Resource Table device (Updated via capsule-on-disk)
│ │   Current version:    16844160
│ │   Minimum Version:    16844160
│ │   Vendor:             Micro-Star International Co., Ltd. (DMI:3mdeb)
│ │   Update State:       Success
│ │   GUID:               119059ec-1a5e-4844-b2e6-6f573b257570
│ │   Device Flags:       • Internal device
│ │                       • Updatable
│ │                       • System requires external power source
│ │                       • Needs a reboot after installation
│ │                       • Cryptographic hash verification is available
│ │                       • Device is usable for the duration of the update
│ │   Device Requests:    • Message
```

and loading the capsule via `fwupdtool install-blob capsule.cap` and selecting System Firmware does trigger an update via CoD. The update finishes without issues. So everything appears to be working correctly

mkopec (approved):

SergiiDmytruk:
> Did you manually place the capsule where needed and set the EFI variable by hand, or did you use some sort of existing tool?

I did it using `CapsuleApp.efi update.cap -OD` in a UEFI shell and also manually.

PR #288, opened 2026-01-14
Actualize UI setting docs

SergiiDmytruk:
Also get rid of dependabot's config (it can't be disabled in any other way) and PR's template from upstream.


macpijan (approved):

macpijan:
Thanks!

PR #292, opened 2026-01-15
Report firmware update result to the user

SergiiDmytruk:
This amends capsule update code to collect some fairly high-level information about how things went, still this is more than one gets from capsule result variables and possibly even logs.  The code takes into account possibility of multiple capsules (we don't know what a user of `fwupd` can do) and multiple payloads.

The displaying part reuses the dialog implemented in https://github.com/Dasharo/edk2/pull/285 after making it generally available and also provides code for constructing dialogs which are then displayed as both text (for serial) and graphics (overrides the text on a display), so that both tests and users should be relatively happy.  In the future, we could reuse this code for our other dialogs and have all of them look consistently.

Screenshots are available in https://github.com/Dasharo/dasharo-issues/issues/1434#issuecomment-3752142245.

coreboot PR: https://github.com/Dasharo/coreboot/pull/826
issue: https://github.com/Dasharo/dasharo-issues/issues/1434
*ref: dsh-1127*

### DIFF
--- a/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
+++ b/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
@@ -915,6 +915,34 @@ GetFmpImageInfoDescriptorVer (
 }
 
 /**
+  Gets pointer to the image data from payload's header.
+
+  @param[in]  ImageHeader   The payload image header.
+
+  @return The pointer to the start of the image data.
+**/
+VOID *
+GetFmpImage (
+  IN EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER  *ImageHeader
+  )
+{
+  if (ImageHeader->Version >= EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER_INIT_VERSION) {
+    return (UINT8 *)(ImageHeader + 1);
+  }
+
+  //
+  // If the EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER is version 1,
+  // Header should exclude UpdateHardwareInstance field, and
+  // ImageCapsuleSupport field if version is 2.
+  //
+  if (ImageHeader->Version == 1) {
+    return (UINT8 *)ImageHeader + OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, UpdateHardwareInstance);
+  }
+
+  return (UINT8 *)ImageHeader + OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, ImageCapsuleSupport);
+}
+
+/**
   Set FMP image data.
 
   @param[in]  Handle        A FMP handle.
@@ -959,20 +987,7 @@ SetFmpImageData (
     mFmpProgress = NULL;
   }
 
-  if (ImageHeader->Version >= EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER_INIT_VERSION) {
-    Image = (UINT8 *)(ImageHeader + 1);
-  } else {
-    //
-    // If the EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER is version 1,
-    // Header should exclude UpdateHardwareInstance field, and
-    // ImageCapsuleSupport field if version is 2.
-    //
-    if (ImageHeader->Version == 1) {
-      Image = (UINT8 *)ImageHeader + OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, UpdateHardwareInstance);
-    } else {
-      Image = (UINT8 *)ImageHeader + OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, ImageCapsuleSupport);
-    }
-  }
+  Image = GetFmpImage (ImageHeader);
 
   if (ImageHeader->UpdateVendorCodeSize == 0) {
     VendorCode = NULL;
--- a/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
+++ b/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
@@ -42,6 +42,8 @@
 #include <Protocol/FirmwareManagementProgress.h>
 #include <Protocol/DevicePath.h>
 
+#include "UpdateReport.h"
+
 EFI_SYSTEM_RESOURCE_TABLE  *mEsrtTable = NULL;
 
 BOOLEAN    mDxeCapsuleLibEndOfDxe      = FALSE;
@@ -1190,9 +1192,10 @@ RecordFmpCapsuleStatus (
 
   This function need support nested FMP capsule.
 
-  @param[in]  CapsuleHeader         Points to a capsule header.
-  @param[in]  CapFileName           Capsule file name.
-  @param[out] ResetRequired         Indicates whether reset is required or not.
+  @param[in]     CapsuleHeader      Points to a capsule header.
+  @param[in]     CapFileName        Capsule file name.
+  @param[out]    ResetRequired      Indicates whether reset is required or not.
+  @param[in,out] CapsuleResult      Outcome of processing the capsule.
 
   @retval EFI_SUCESS            Process Capsule Image successfully.
   @retval EFI_UNSUPPORTED       Capsule image is not supported by the firmware.
@@ -1200,11 +1203,13 @@ RecordFmpCapsuleStatus (
   @retval EFI_OUT_OF_RESOURCES  Not enough memory.
   @retval EFI_NOT_READY         No FMP protocol to handle this FMP capsule.
 **/
+STATIC
 EFI_STATUS
 ProcessFmpCapsuleImage (
   IN EFI_CAPSULE_HEADER  *CapsuleHeader,
   IN CHAR16              *CapFileName   OPTIONAL,
-  OUT BOOLEAN            *ResetRequired OPTIONAL
+  OUT BOOLEAN            *ResetRequired OPTIONAL,
+  IN OUT CapsuleResult   *CapsuleResult OPTIONAL
   )
 {
   EFI_STATUS                                    Status;
@@ -1221,9 +1226,10 @@ ProcessFmpCapsuleImage (
   UINTN                                         Index2;
   BOOLEAN                                       NotReady;
   BOOLEAN                                       Abort;
+  BOOLEAN                                       PayloadAborted;
 
   if (!IsFmpCapsuleGuid (&CapsuleHeader->CapsuleGuid)) {
-    return ProcessFmpCapsuleImage ((EFI_CAPSULE_HEADER *)((UINTN)CapsuleHeader + CapsuleHeader->HeaderSize), CapFileName, ResetRequired);
+    return ProcessFmpCapsuleImage ((EFI_CAPSULE_HEADER *)((UINTN)CapsuleHeader + CapsuleHeader->HeaderSize), CapFileName, ResetRequired, CapsuleResult);
   }
 
   NotReady = FALSE;
@@ -1269,6 +1275,7 @@ ProcessFmpCapsuleImage (
                );
     if (EFI_ERROR (Status)) {
       DEBUG ((DEBUG_ERROR, "Driver Return Status = %r\n", Status));
+      ReportCapsuleOutcome (CapsuleResult, CAPSULE_DRIVER, Status);
       return Status;
     }
   }
@@ -1317,6 +1324,7 @@ ProcessFmpCapsuleImage (
       continue;
     }
 
+    PayloadAborted = FALSE;
     for (Index2 = 0; Index2 < NumberOfHandles; Index2++) {
       if (Abort) {
         RecordFmpCapsuleStatus (
@@ -1337,6 +1345,7 @@ ProcessFmpCapsuleImage (
                  );
       if (Status != EFI_SUCCESS) {
         Abort = TRUE;
+        PayloadAborted = TRUE;
       } else {
         if (ResetRequired != NULL) {
           *ResetRequired |= ResetRequiredBuffer[Index2];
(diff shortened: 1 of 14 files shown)


MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+  PopUpInit (&PopUp, /*Width=*/78);
+
+  if (Report->CapsuleCount == 0) {
+    Success = FALSE;
+    AddTitle (&PopUp, L"Firmware Update Skipped");
+    AddLine (&PopUp, L"The firmware is in capsule update mode, yet no update capsules were");
+    AddLine (&PopUp, L"discovered. The update was cancelled without modifying the firmware.");
+    AddLine (&PopUp, L"");
+    AddLine (&PopUp, L"However, because this is not normal, consider contacting us at");
+    AddLine (&PopUp, L"support@dasharo.com describing how this situation came about.");
```
BeataZdunczyk:
```suggestion
    AddLine (&PopUp, L"support@dasharo.com describing how this situation occurred.");
```

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+
+  PopUpInit (&PopUp, /*Width=*/78);
+
+  if (Report->CapsuleCount == 0) {
+    Success = FALSE;
+    AddTitle (&PopUp, L"Firmware Update Skipped");
+    AddLine (&PopUp, L"The firmware is in capsule update mode, yet no update capsules were");
+    AddLine (&PopUp, L"discovered. The update was cancelled without modifying the firmware.");
+    AddLine (&PopUp, L"");
+    AddLine (&PopUp, L"However, because this is not normal, consider contacting us at");
```
BeataZdunczyk: I propose something like:

```suggestion
    AddLine (&PopUp, L"This situation is unexpected. Please consider contacting us at");
```

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+      AddLineF (&PopUp, L"Successfully applied all firmware updates (%d capsules).", Report->CapsuleCount);
+    }
+  } else {
+    Success = FALSE;
+    AddTitle (&PopUp, L"Firmware Update Failed");
+    if (Report->CapsuleCount == 1) {
+      AddLine (&PopUp, L"A firmware update was not applied successfully. Depending on the severity of");
+      AddLine (&PopUp, L"the error the device may be unbootable and require external firmware flashing.");
+    } else {
+      AddLine (&PopUp, L"At least one firmware update was not applied successfully. Depending on the");
```
BeataZdunczyk: > At least one firmware update was not applied successfully.

Is the distinction in “The firmware update was not applied successfully” meaningful for users? Would “The firmware update was not applied successfully” be clearer?

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+      AddLine (&PopUp, L"Successfully applied a firmware update.");
+    } else {
+      AddLineF (&PopUp, L"Successfully applied all firmware updates (%d capsules).", Report->CapsuleCount);
+    }
+  } else {
+    Success = FALSE;
+    AddTitle (&PopUp, L"Firmware Update Failed");
+    if (Report->CapsuleCount == 1) {
+      AddLine (&PopUp, L"A firmware update was not applied successfully. Depending on the severity of");
+      AddLine (&PopUp, L"the error the device may be unbootable and require external firmware flashing.");
```
BeataZdunczyk: The current wording feels alarming.

> may be unbootable

Are we certain about this? If this is not a confirmed state, the wording introduces a lot of anxiety for users.

If this scenario is possible, we should also provide users with clear, actionable guidance. In addition to asking them to contact us, we could point them to docs.dasharo.com, where we have recovery instructions. This would give users a clear recovery path if needed.

Something like: `Please refer to the device-specific recovery instructions at docs.dasharo.com.` would be suffitient.

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+    AddLine (&PopUp, L"");
+    AddLine (&PopUp, L"However, because this is not normal, consider contacting us at");
+    AddLine (&PopUp, L"support@dasharo.com describing how this situation came about.");
+  } else if (AllUpdatesSuccessful (Report)) {
+    Success = TRUE;
+    AddTitle (&PopUp, L"Firmware Update Succeeded");
+    if (Report->CapsuleCount == 1) {
+      AddLine (&PopUp, L"Successfully applied a firmware update.");
+    } else {
+      AddLineF (&PopUp, L"Successfully applied all firmware updates (%d capsules).", Report->CapsuleCount);
```
filipleple, line 615: The popup looks good, but in the issue I see 
> Must clearly indicate the outcome of an update: either a success (with information about old and new versions) 

So it looks like we want to add the version information

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+VOID
+EFIAPI
+ReportAddPayload (
+  IN OUT CapsuleResult  *Capsule OPTIONAL,
+  IN EFI_STATUS         Status
+  )
+{
+  VOID           *Payloads;
+  PayloadResult  *PayloadResult;
+
```
filipleple, line 169: I think we need a `if (Capsule == NULL) return;` immediately if `Capsule` is `OPTIONAL`

MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
```diff
@@ -1362,6 +1382,10 @@ ProcessFmpCapsuleImage (
     }
   }
 
+  if (Abort) {
+    ReportCapsuleOutcome (CapsuleResult, CAPSULE_PAYLOAD, EFI_SUCCESS);
```
filipleple: if Abort then probably:

```suggestion
    ReportCapsuleOutcome (CapsuleResult, CAPSULE_PAYLOAD, EFI_ABORTED);
```

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  Report->Capsules = Capsules;
+
+  CapsuleResult = &Report->Capsules[Report->CapsuleCount++];
+
+  CapsuleResult->Index   = Index;
+  CapsuleResult->Outcome = CAPSULE_UNKNOWN;
+  CapsuleResult->Status  = EFI_SUCCESS;
+
```
filipleple, line 115: I think we might want to initialize the remaining `Payloads` and `PayloadCount` fields as well here for safety, so that e.g. a garbage pointer doesn't go into `FreePool`, and we don't crash on `for (PayloadIdx = 0; PayloadIdx < Capsule->PayloadCount; ++PayloadIdx) { ... }`

filipleple (changes requested):

BeataZdunczyk (changes requested):

MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
SergiiDmytruk: I probably meant to suppres printing status in this case (`EFI_SUCCESS` isn't reported), but will use `EFI_ABORTED` as it's not clear where the status is synthetic and is worth omitting.

https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-cea14f6f2ac424721e02fc285d6da696d786acf464bb0c5c982271012347a7f2R1401

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: It becomes less obvious after taking arbitrary capsules into account, but I've added versions now in a somewhat reasonable way (turned out to be quite involved to get to them).  See 4cea593edd4004fb252dab13f9a232c2c34faa72.

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: Forgot about it, thanks.

https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-fbc1ef542448c79a2b70a1093504b95e18f7320f3a5b17a240813f8767f8ef3cR170-R173

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: Unlike with `realloc()` new memory is  zero-initialized by `ReallocatePool()`:

```c
/**
  Reallocates a buffer of type EfiBootServicesData.

  Allocates and zeros the number bytes specified by NewSize from memory of type
  EfiBootServicesData.  If OldBuffer is not NULL, then the smaller of OldSize and
  ...
```

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: This is for the unlikely case of multiple capsules when at least one has failed, the list of capsules should provide the context for the message.  Simplifying the text should not make it less accurate.  Rephrased as `Not all firmware updates were applied successfully.`.

https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-fbc1ef542448c79a2b70a1093504b95e18f7320f3a5b17a240813f8767f8ef3cR623
https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-fbc1ef542448c79a2b70a1093504b95e18f7320f3a5b17a240813f8767f8ef3cR632

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: It should be alarming if system's flash in an unknown state, which is the state after a failed update unless proven otherwise.  https://github.com/Dasharo/dasharo-issues/issues/1435 can mitigate this, but won't necessarily be able to communicate that the firmware is in a reasonable state (depends on how the implementation will go).  Changed the message when the firmware is expected to remain functional: https://github.com/Dasharo/edk2/pull/292/commits/d315e564b5a15e44c4f8b643bf2c8be79849c383

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-fbc1ef542448c79a2b70a1093504b95e18f7320f3a5b17a240813f8767f8ef3cR605

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-fbc1ef542448c79a2b70a1093504b95e18f7320f3a5b17a240813f8767f8ef3cR604

SergiiDmytruk (commented):
In addition to the review comments:
 - fixed [a memory bug](https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-1251b83d81a12b025e45c6866fee30549b000a825337436e5b816e2c5cbde3e0)
 - added reporting capsule result in [one more place](https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-cea14f6f2ac424721e02fc285d6da696d786acf464bb0c5c982271012347a7f2R1627)
 - added displaying [capsule GUIDs](https://github.com/Dasharo/edk2/pull/292/commits/08333264dff5cdbc534007998a8c47ffc9fdaeab)
 - increased delay on success [from 5 to 30 seconds](https://github.com/Dasharo/edk2/compare/3cf77732acb2ff06dae49088ce22d9e796855d0f..08333264dff5cdbc534007998a8c47ffc9fdaeab#diff-fbc1ef542448c79a2b70a1093504b95e18f7320f3a5b17a240813f8767f8ef3cR648-R655)

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
filipleple: Got it, thanks

filipleple (approved):
LGTM now 👍

SergiiDmytruk:
I meant to include new screenshots yesterday, but must have closed that tab before sending them.  Updated look, this time next to serial output (`v0.1.1-rc1` is the expected fake version):

image

***

image

***

image

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
BeataZdunczyk: ok, got it, thank you

BeataZdunczyk:
image


it’s unclear which firmware version is currently runnig on a hardware.

@SergiiDmytruk Instead of `running firmware` maybe we should refer to it as the **previous** firmware version?

mkopec:
Maybe simply `Updated from version x to y`? :)

SergiiDmytruk:
@BeataZdunczyk, [updated](https://github.com/Dasharo/edk2/compare/08333264dff5cdbc534007998a8c47ffc9fdaeab..c2e30de8c510567f520f99d6c62b20964349afe3):
image

BeataZdunczyk:
> @BeataZdunczyk, [updated](https://github.com/Dasharo/edk2/compare/08333264dff5cdbc534007998a8c47ffc9fdaeab..c2e30de8c510567f520f99d6c62b20964349afe3):
image

@SergiiDmytruk , perfect, approved.

BeataZdunczyk (approved):

PR #295, opened 2026-01-30
Improve capsule update progress bar

filipleple:
issue: https://github.com/Dasharo/dasharo-issues/issues/1439
coreboot PR: https://github.com/Dasharo/coreboot/pull/837
_ref: DSH-1131_

### DIFF
--- a/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
+++ b/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
@@ -616,6 +616,36 @@ MigrateBootLogo (
 
 STATIC
 BOOLEAN
+MigrateMrcCache (
+  IN CONST MigrationData  *Data
+  )
+{
+  RegionMigrationStatus  Status;
+
+  Status = MigrateRegion ("RW_MRC_CACHE", Data, FALSE);
+
+  return Status == REGION_MIGRATED
+      || Status == REGION_NOT_IN_SRC
+      || Status == REGION_NOT_IN_DST;
+}
+
+STATIC
+BOOLEAN
+MigrateRecoveryMrcCache (
+  IN CONST MigrationData  *Data
+  )
+{
+  RegionMigrationStatus  Status;
+
+  Status = MigrateRegion ("RECOVERY_MRC_CACHE", Data, FALSE);
+
+  return Status == REGION_MIGRATED
+      || Status == REGION_NOT_IN_SRC
+      || Status == REGION_NOT_IN_DST;
+}
+
+STATIC
+BOOLEAN
 MigrateGbeRegion (
   IN CONST MigrationData  *Data
   )
@@ -1000,6 +1030,16 @@ MergeFirmwareImages (
 
   if (!MigrateGbeRegion (&Data)) {
     DEBUG ((DEBUG_ERROR, "%a(): MigrateGbeRegion () failed\n", __FUNCTION__));
+    goto Fail;
+  }
+
+  if (!MigrateMrcCache (&Data)) {
+    DEBUG ((DEBUG_ERROR, "%a(): MigrateMrcCache () failed\n", __FUNCTION__));
+    goto Fail;
+  }
+
+  if (!MigrateRecoveryMrcCache (&Data)) {
+    DEBUG ((DEBUG_ERROR, "%a(): MigrateMrcCache () failed\n", __FUNCTION__));
     goto Fail;
   }
 
--- a/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
+++ b/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
@@ -1053,7 +1053,7 @@ MergeFirmwareImages (
   }
 
   if (!MigrateRecoveryMrcCache (&Data)) {
-    DEBUG ((DEBUG_ERROR, "%a(): MigrateMrcCache () failed\n", __FUNCTION__));
+    DEBUG ((DEBUG_ERROR, "%a(): MigrateRecoveryMrcCache () failed\n", __FUNCTION__));
     goto Fail;
   }
 
--- a/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
+++ b/DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
@@ -200,14 +200,21 @@ GetFmap (
 }
 
 /**
-  Read current firmware in full and return as newly allocated pool memory.
+  Read current firmware in full and return as newly allocated pool memory,
+  with optional progress reporting via callback.
+
+  @param[in] IsDescriptorLocked   Whether the flash descriptor is locked.
+  @param[in] ProgressCallback     Optional callback invoked after each block read.
+  @param[in] ProgressContext      Caller-provided context passed to callback.
 
   @return NULL  On error.
 **/
 VOID *
 EFIAPI
 ReadCurrentFirmware (
-  BOOLEAN IsDescriptorLocked
+  IN BOOLEAN                    IsDescriptorLocked,
+  IN FW_READ_PROGRESS_CALLBACK  ProgressCallback OPTIONAL,
+  IN VOID                       *ProgressContext OPTIONAL
   )
 {
   EFI_STATUS  Status;
@@ -216,6 +223,7 @@ ReadCurrentFirmware (
   UINTN       Block;
   UINTN       BlockSize;
   UINTN       NumBytes;
+  UINTN       BlockCount;
 
   Status = SmmStoreLibGetBlockSize (&BlockSize);
   if (EFI_ERROR (Status)) {
@@ -251,7 +259,9 @@ ReadCurrentFirmware (
     return NULL;
   }
 
-  for (Block = 0; Block < FwSize / BlockSize; Block++) {
+  BlockCount = FwSize / BlockSize;
+
+  for (Block = 0; Block < BlockCount; Block++) {
     NumBytes = BlockSize;
     Status = SmmStoreLibReadAnyBlock (
                Block,
@@ -278,6 +288,10 @@ ReadCurrentFirmware (
         FreePool (Image);
         return NULL;
       }
+    }
+
+    if (ProgressCallback != NULL) {
+      ProgressCallback (Block, BlockCount, ProgressContext);
     }
   }
 
--- a/DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
+++ b/DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
@@ -826,7 +826,7 @@ CountDifferingBlocks (
       continue;
     }
 
-    if (!IsRangeWriteable (Image2, Size, Offset, BlockSize)) {
+    if (!IsRangeWriteable (Image1, Size, Offset, BlockSize)) {
       continue;
     }
 
@@ -1248,6 +1248,7 @@ FmpDeviceSetImageWithStatus (
   ProgressContext  ProgressCtx;
   UINTN            DifferingBlocks;
   UINTN            WrittenBlocks;
+  UINTN            ScaledWriteWeight;
   BOOLEAN          SmoothProgress;
 
   *LastAttemptStatus = LAST_ATTEMPT_STATUS_ERROR_UNSUCCESSFUL;
@@ -1340,6 +1341,7 @@ FmpDeviceSetImageWithStatus (
   //
   // Phase 2: Write differing blocks.
   //
+  ScaledWriteWeight = UPDATE_BLOCK_WEIGHT;
   if (SmoothProgress) {
     //
     // Count how many blocks actually differ to get accurate progress.
@@ -1355,10 +1357,13 @@ FmpDeviceSetImageWithStatus (
             __FUNCTION__, DifferingBlocks, BlockCount));
 
     //
-    // Recalculate TotalUnits based on actual differing blocks.
-    // CurrentUnits already has the read progress (BlockCount * READ_BLOCK_WEIGHT).
+    // Scale the per-block write weight so that writing all differing blocks
+    // still consumes exactly BlockCount * UPDATE_BLOCK_WEIGHT units, keeping
+    // TotalUnits unchanged and avoiding a sudden progress jump.
     //
-    ProgressCtx.TotalUnits = BlockCount * READ_BLOCK_WEIGHT + DifferingBlocks * UPDATE_BLOCK_WEIGHT;
+    if (DifferingBlocks > 0) {
+      ScaledWriteWeight = UPDATE_BLOCK_WEIGHT * BlockCount / DifferingBlocks;
+    }
   }
 
   Offset = 0;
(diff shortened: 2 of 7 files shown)


DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -277,11 +287,29 @@ ReadCurrentFirmware (
...
+  @return NULL  On error.
+**/
+VOID *
+EFIAPI
+ReadCurrentFirmware (
+  BOOLEAN IsDescriptorLocked
+  )
+{
+  return ReadCurrentFirmwareWithProgress (IsDescriptorLocked, NULL, NULL);
+}
```
SergiiDmytruk: Why preserve the original function?  It was called from a single place, you could just update it in place.  Callback being optional is actually useful for recovery which will read the firmware without reporting progress.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1125,24 +1235,38 @@ FmpDeviceSetImageWithStatus (
         "%a(): range %d - %d is not writeable\n",
         __FUNCTION__, Offset, BlockSize + Offset
         ));
-      IncrementProgress (Progress, TotalSteps, 2, &Step, &ShouldReportProgress);
+      //
+      // Block is not writeable - skip without incrementing progress.
+      //
```
SergiiDmytruk: But if we got here, the updated block is different from the original and is counted in `ProgressCtx.TotalUnits`.

SergiiDmytruk (changes requested):
The new functionality should be guarded by a PCD (assuming that all blocks differ when it's off should be good enough).  And coreboot should enable it, so likely the PRs need to be based on https://github.com/Dasharo/coreboot/pull/836 and https://github.com/Dasharo/edk2/pull/294 as those two change how new capsule features are controlled.

filipleple:
OK, marking as draft and addressing the review

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
filipleple: [Updated](https://github.com/Dasharo/edk2/pull/295/changes#diff-df48c35c52e8b3221b927ccc9e290725c37320ea2077f5b0862609923f3a8aa0R214)

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: [Fixed](https://github.com/Dasharo/edk2/pull/295/changes#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fR1252)

filipleple (commented):
Addressed the review

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -612,6 +626,21 @@ MigrateBootLogo (
...
 
+STATIC
+BOOLEAN
+MigrateMrcCache (
+  IN CONST MigrationData  *Data
+  )
+{
+  RegionMigrationStatus  Status;
+
+  Status = MigrateRegion ("RW_MRC_CACHE", Data, FALSE);
```
SergiiDmytruk, line 639: Is there any point in copying `RECOVERY_MRC_CACHE` as well?  Not sure when it gets used, but we have it in FMAPs and it seems to be enabled in configs.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -693,42 +694,47 @@ FmpDeviceSetImage (
            );
 }
 
+//
+// Weight constants for progress calculation.
+// Reading 8 blocks takes approximately the same time as 1 erase+write operation.
+//
+#define READ_BLOCK_WEIGHT   1
+#define WRITE_BLOCK_WEIGHT  8
```
SergiiDmytruk: Maybe rename this to `UPDATE_BLOCK_WEIGHT`?  I had to  remind myself that it also accounts for erase and not just write.

MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
```diff
@@ -370,23 +387,118 @@ DisplayUpdateProgress (
...
 
   //
-  // Make sure a valid start, end, and size info are available (find the Logo)
+  // Text fallback rendering when GOP is unavailable
   //
-  if (!mGraphicsGood) {
-    DEBUG ((DEBUG_INFO, "Graphics Not Good.  Not doing any onscreen visual display\n"));
-    return EFI_NOT_READY;
+  if (mTextFallback) {
+    UINTN  CurrentAttribute;
```
SergiiDmytruk: I would move body of this `if` to a separate function not least to potentially have easier time during a rebase.

MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
SergiiDmytruk, line 1: I think these changes obsolete `GRAPHICAL_CAPSULE_PROGRESS` macro in `DasharoPaloadPkg.dsc` and `CONFIG_EDK2_GRAPHICAL_CAPSULE_PROGRESS` in coreboot.  We can always use this library now.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
SergiiDmytruk: This works, but `SI_DESC`, `SI_ME`, `RW_UNUSED` and `SI_GBE` can account for half of a flash chip.  We should probably exclude protected ranges from diffing and not report here with enabled `PcdSmoothCapsuleProgress` to not compromise the improvements.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -1103,15 +1201,40 @@ FmpDeviceSetImageWithStatus (
...
+    //
+    DifferingBlocks = CountDifferingBlocks (CurrentImage, UpdatedImage, ImageSize, BlockSize);
+    DEBUG ((DEBUG_INFO, "%a(): %d blocks differ out of %d total\n",
+            __FUNCTION__, DifferingBlocks, BlockCount));
+
+    //
+    // Recalculate TotalUnits based on actual differing blocks.
+    // CurrentUnits already has the read progress (BlockCount * READ_BLOCK_WEIGHT).
+    //
+    ProgressCtx.TotalUnits = BlockCount * READ_BLOCK_WEIGHT + DifferingBlocks * WRITE_BLOCK_WEIGHT;
```
SergiiDmytruk: I've been thinking about this line: the smaller the number of changed blocks, the bigger the jump here.

How about keeping total number of units the same but scaling the weight?  I.e. using `WRITE_BLOCK_WEIGHT * BlockCount / DifferingBlocks` which still adds up to `WRITE_BLOCK_WEIGHT * BlockCount` (maybe not quite due to integer division, but the progress is forced to 100% in FmpDxe after calling this function anyway).  This way the jump will be distributed evenly and become less noticeable.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
filipleple: Yup, it's RAM training cache, if we copy it we don't have to wait for RAM re-training after each update (see https://github.com/Dasharo/dasharo-issues/issues/1747).

I did it on the side and left it in accidentally, but when I noticed the speedup when testing I thought we might actually include it in this PR already

SergiiDmytruk: I didn't expect to see it implemented in this PR, but see no downsides.  My point is that the code copies only one of two caches:
```fmap
   UNIFIED_MRC_CACHE(PRESERVE) 128K {
    RECOVERY_MRC_CACHE 64K
    RW_MRC_CACHE 64K
   }
```

filipleple: Oh yeah right, both should be migrated, [fixed](https://github.com/Dasharo/edk2/pull/295/changes/7e57d219fd3d3c07f4073dd0d21321c032641b8e#diff-df48c35c52e8b3221b927ccc9e290725c37320ea2077f5b0862609923f3a8aa0R634)

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: [Done](https://github.com/Dasharo/edk2/pull/295/changes#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fR706)

MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
filipleple: [Moved](https://github.com/Dasharo/edk2/pull/295/changes#diff-fd71722f2cf2bd449229bd5c30da7f61785a88fed87be8887bc846e2a4889b9fR333), made it into a static within the same file for now, not sure if moving it out to another file would help

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: [removed](https://github.com/Dasharo/edk2/pull/295/changes#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fR1362) and progress now advances only on actual write work since protected regions are [excluded](https://github.com/Dasharo/edk2/pull/295/changes#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fR829)

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: Right, [exlcluded](https://github.com/Dasharo/edk2/pull/295/changes#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fR829)

MdeModulePkg/Library/DisplayUpdateProgressLibGraphics/DisplayUpdateProgressLibGraphics.c
filipleple: [Removed](https://github.com/Dasharo/edk2/pull/295/changes/0046a2387d120316dffdf6fbdbe9c0aa1f86a077), in coreboot as well: https://github.com/Dasharo/coreboot/pull/837/changes/6658deca7cc2703a1ed533d833e7e0403a67b9f4

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
```diff
@@ -1003,6 +1047,16 @@ MergeFirmwareImages (
     goto Fail;
   }
 
+  if (!MigrateMrcCache (&Data)) {
+    DEBUG ((DEBUG_ERROR, "%a(): MigrateMrcCache () failed\n", __FUNCTION__));
+    goto Fail;
+  }
+
+  if (!MigrateRecoveryMrcCache (&Data)) {
+    DEBUG ((DEBUG_ERROR, "%a(): MigrateMrcCache () failed\n", __FUNCTION__));
```
SergiiDmytruk:
```suggestion
    DEBUG ((DEBUG_ERROR, "%a(): MigrateRecoveryMrcCache () failed\n", __FUNCTION__));
```

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
```diff
@@ -742,12 +747,93 @@ IncrementProgress (
...
+  BlockCount = Size / BlockSize;
+  DifferingCount = 0;
+  Offset = 0;
+
+  for (Block = 0; Block < BlockCount; Block++, Offset += BlockSize) {
+    if (CompareMem (Image1 + Offset, Image2 + Offset, BlockSize) == 0) {
+      continue;
+    }
+
+    if (!IsRangeWriteable (Image2, Size, Offset, BlockSize)) {
```
SergiiDmytruk: `Image2` is `UpdatedImage`.  I think that what's writable at the moment is defined by the current firmware (although both should match in most cases).

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
SergiiDmytruk: Now there is no need for
```c
      ProgressCtx.CurrentUnits += UPDATE_BLOCK_WEIGHT;
      ReportProgress (&ProgressCtx);
```
below.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
SergiiDmytruk: Not doing `ReportProgress (&ProgressCtx);` right away merely postpones the jump.  If you have 32 MiB flash, then:
 - there are `32768/64 = 512` blocks
 - initial value of `TotalUnits` is `(512*1) + (512*8) = 4608`
 - if only quoter of the flash is updated, the new `TotalUnits` is  `(512*1) + (128*8) = 1536`
 - `CurrentUnits` after reading is `512` which means that `512/4608 = 11%` becomes `512/1536 = 33%` on the first write

This `11% -> 33%` change is what I'm talking about.

DasharoPayloadPkg/Library/FmpDeviceSmmLib/Flashing.c
filipleple: https://github.com/Dasharo/edk2/pull/295/changes/2664b4f12f13e2315449e63ef5c21c6537ac3b66#diff-df48c35c52e8b3221b927ccc9e290725c37320ea2077f5b0862609923f3a8aa0R1056

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: https://github.com/Dasharo/edk2/pull/295/changes/2664b4f12f13e2315449e63ef5c21c6537ac3b66#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fR829

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: https://github.com/Dasharo/edk2/pull/295/changes/2664b4f12f13e2315449e63ef5c21c6537ac3b66#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fL1392 I think this should work, the progress between 11 and 33 looks smoother

https://github.com/user-attachments/assets/c4f83163-5150-4b36-9c59-186226a0c033

DasharoPayloadPkg/Library/FmpDeviceSmmLib/FmpDeviceSmmLib.c
filipleple: https://github.com/Dasharo/edk2/pull/295/changes/2664b4f12f13e2315449e63ef5c21c6537ac3b66#diff-b33849bc60c6d8a499fabcaf6d579fa92740f3e0e5b0ebb22c24db8af56d307fL1395

SergiiDmytruk (approved):

PR #296, opened 2026-01-30
FmpDeviceSmmLib/Flashing.c: migrate MRC cache across capsule updates

filipleple:
MRC cache was being cleared on every capsule update, forcing a full memory retraining cycle on the next boot. This is unnecessary because the cache is versioned by FSP and will be invalidated automatically if the FSP binary changes in the new firmware.

Migrate the RW_MRC_CACHE FMAP region the same way ROMHOLE, BOOTSPLASH and SI_GBE are already handled. The region is not offset-sensitive and its absence in either image is non-fatal.


SergiiDmytruk:
A more completion version of this change has been merged as part of https://github.com/Dasharo/edk2/pull/295, so closing.

PR #297, opened 2026-02-03
Capsules v2: recovery

SergiiDmytruk:
* Fixed a couple bugs, see commits.
* Added wrapping long lines to `PopUpLib`.
* Added storing abort reason from an FMP capsules to update report.  It shows up in a payload status as shown below.  "the device may be unbootable" near the top still makes sense because we can't exclude this possibility (e.g., if ME region was overwritten while ME is running, returning it to its initial state doesn't necessarily mean everything is fine because by that time ME's runtime state may have changed in an incompatible way).
* During recovery each modified block is restored up to 3 times before moving on to the next one.  Recovery skips write-protected regions and part of the flash which wasn't affected by the update before hitting an error.

image

Elements of the following patch could be used to simulate an update failure:
```patch
 src/drivers/smmstore/store.c | 16 ++++++++++++++++
 1 file changed, 16 insertions(+)

diff --git a/src/drivers/smmstore/store.c b/src/drivers/smmstore/store.c
index 2fca79f180..4b794c72e1 100644
--- a/src/drivers/smmstore/store.c
+++ b/src/drivers/smmstore/store.c
@@ -8,6 +8,7 @@
 #include <fmap.h>
 #include <fmap_config.h>
 #include <smmstore.h>
+#include <string.h>
 #include <types.h>
 
 #define SMMSTORE_REGION "SMMSTORE"
@@ -440,6 +441,13 @@ int smmstore_rawread_region(uint32_t block_id, uint32_t offset, uint32_t bufsize
 	if (!ptr)
 		return -1;
 
+	// Make reads return inconsistent data to "fail" recovery.
+	/* static int value; */
+	/* if (use_full_flash && block_id == 1) { */
+	/* 	memset(ptr, value++, bufsize); */
+	/* 	return 0; */
+	/* } */
+
 	printk(BIOS_DEBUG, "smm store: reading %p block %d, offset=0x%x, size=%x\n",
 	       ptr, block_id, offset, bufsize);
 
@@ -467,6 +475,10 @@ int smmstore_rawwrite_region(uint32_t block_id, uint32_t offset, uint32_t bufsiz
 	struct region_device store;
 	struct region_device com_buf;
 
+	// Simulate persistent write failure.
+	if (use_full_flash && block_id == 1)
+		return -1;
+
 	if (lookup_block_in_store(&store, block_id) < 0)
 		return -1;
 
@@ -501,6 +513,10 @@ int smmstore_rawclear_region(uint32_t block_id)
 {
 	struct region_device store;
 
+	// Simulate persistent erase failure.
+	/* if (use_full_flash && block_id == 1) */
+	/* 	return -1; */
+
 	if (lookup_block_in_store(&store, block_id) < 0)
 		return -1;
 
```

issue: https://github.com/Dasharo/dasharo-issues/issues/1435
coreboot PR: https://github.com/Dasharo/coreboot/pull/838
*ref: dsh-1128*

### DIFF
--- a/MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
+++ b/MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
@@ -72,10 +72,21 @@ ReportFree (
   )
 {
   UINTN          CapsuleIdx;
+  UINTN          PayloadIdx;
   CapsuleResult  *Capsule;
+  PayloadResult  *Payload;
 
   for (CapsuleIdx = 0; CapsuleIdx < Report->CapsuleCount; ++CapsuleIdx) {
     Capsule = &Report->Capsules[CapsuleIdx];
+
+    for (PayloadIdx = 0; PayloadIdx < Capsule->PayloadCount; ++PayloadIdx) {
+      Payload = &Capsule->Payloads[PayloadIdx];
+
+      if (Payload->Message != NULL) {
+        FreePool (Payload->Message);
+      }
+    }
+
     if (Capsule->Payloads != NULL) {
       FreePool (Capsule->Payloads);
     }
@@ -161,7 +172,8 @@ VOID
 EFIAPI
 ReportAddPayload (
   IN OUT CapsuleResult  *Capsule OPTIONAL,
-  IN EFI_STATUS         Status
+  IN EFI_STATUS         Status,
+  IN CONST CHAR16       *Message
   )
 {
   VOID           *Payloads;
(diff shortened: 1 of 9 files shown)


MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
```diff
@@ -183,6 +195,14 @@ ReportAddPayload (
   PayloadResult = &Capsule->Payloads[Capsule->PayloadCount++];
 
   PayloadResult->Status = Status;
+
```
filipleple, line 198: @SergiiDmytruk I think we could init `PayloadResult->Message` as `NULL` here to be safe, unless that's handled elsewhere already

filipleple:
image

Managed to reproduce with the patch from the initial commit

MdeModulePkg/Library/DxeCapsuleLibFmp/UpdateReport.c
SergiiDmytruk: That wouldn't hurt, but the new element, referred to by `PayloadResult`, is added via `ReallocatePool()` that ensures it's zero-initialized.

filipleple: O, okay got it now

filipleple (approved):

PR #301, opened 2026-02-09
Firmware Update Mode: Add separate variable for FUM Request

mkopec:
Fixes an issue where entering FUM twice in a row is prevented by the Request variable being the same as the Active variable, being volatile and without runtime write access.

See: https://github.com/Dasharo/dasharo-issues/issues/1759

TEST=Boot Qemu and confirm FUM may be entered twice in a row. Confirm that resetting the platform still clears FUM.

coreboot PR: https://github.com/Dasharo/coreboot/pull/843
*ref: ncm-2145*


DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
```diff
@@ -571,7 +571,7 @@ DasharoEnableFUM (
 
   Enable = TRUE;
   Status = gRT->SetVariable (
-      DASHARO_VAR_FIRMWARE_UPDATE_MODE,
+      DASHARO_VAR_FIRMWARE_UPDATE_MODE_REQUEST,
       &gDasharoSystemFeaturesGuid,
       EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS | EFI_VARIABLE_NON_VOLATILE,
```
SergiiDmytruk, line 576: `DASHARO_VAR_FIRMWARE_UPDATE_MODE_REQUEST` shouldn't need `EFI_VARIABLE_RUNTIME_ACCESS` attribute.

SergiiDmytruk (commented):
Why did you choose to add another variable instead of just removing non-volatile one before `SetVariable()` call?  I'm wondering whether we can now check for `DASHARO_VAR_FIRMWARE_UPDATE_MODE_REQUEST` after it's gone, previously we would still read the value thanks to the identical variable name.

mkopec:
> Why did you choose to add another variable instead of just removing non-volatile one before SetVariable() call? I'm wondering whether we can now check for DASHARO_VAR_FIRMWARE_UPDATE_MODE_REQUEST after it's gone, previously we would still read the value thanks to the identical variable name.

That would work too, i would just have to give the volatile FUM variable runtime access, to allow runtime capsule update. I believe the runtime FUM var has no runtime access specifically for security, so I didn't want to change that, but I can do that in the next version.

SergiiDmytruk:
> That would work too, i would just have to give the volatile FUM variable runtime access, to allow runtime capsule update. I believe the runtime FUM var has no runtime access specifically for security, so I didn't want to change that, but I can do that in the next version.

But it already has runtime access (this is on `dasharo`): https://github.com/Dasharo/edk2/blob/edbff52d39d1420a22cc4df8b56d8e78dd43fce4/DasharoPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c#L825-L831

In fact the volatile version was added specifically for runtime access in b68acc46a6c, so that iPXE script could test for it.

mkopec:
Applied your suggestion @SergiiDmytruk

mkopec:
Ehh, @SergiiDmytruk that approach won't work for capsule update from Linux:


> Once ExitBootServices() is performed, only variables that have EFI_VARIABLE_RUNTIME_ACCESS and EFI_VARIABLE_NON_VOLATILE set can be set with SetVariable(). Variables that have runtime access but that are not nonvolatile are read-only data variables once ExitBootServices() is performed. 

Looks like it has to be 2 variables, unless you have any other suggestions :\

DasharoModulePkg/Library/DasharoVariablesLib/DasharoVariablesLib.c
SergiiDmytruk: Looks like only variables with `EFI_VARIABLE_RUNTIME_ACCESS` can be created at runtime.

SergiiDmytruk (approved):
I wanted deletion because it was mentioned in https://github.com/Dasharo/dasharo-issues/issues/1759 and we were already doing it in `PromoteFUM()`.  Now we know that we actually need the second variable.

***

Reverted PRs to their previous state.

PR #303, opened 2026-02-12
NVC ADL-P fixes 

filipleple:
Add the Bayhub SD card driver, making SD cards bootable from EDK2. Adresses: https://github.com/Dasharo/dasharo-issues/issues/454

Add a note that the `Keep IOMMU enabled(...)` option is incompatible with Windows _on many devices_, making the OS unable to boot. This is to answer the issues: https://github.com/Dasharo/dasharo-issues/issues/1771 https://github.com/Dasharo/dasharo-issues/issues/641

https://github.com/Dasharo/coreboot/pull/846
_ref: NCM-2103_


miczyg1:
> Add a note that the `Keep IOMMU enabled(...)` option is incompatible with Windows, making the OS unable to boot.

It isn't necessarily incompatible. I have succeeded in booting Windows when this option is enabled on certain devices.

mkopec (approved):

SergiiDmytruk (approved):
Worth noting that `mOverride = &BhtOverride;` suppresses looking up `gEdkiiSdMmcOverrideProtocolGuid` which is provided by platform-specific drivers.  It doesn't matter now since we don't use any alternative SD-card drivers. This just isn't a proper way of integration and probably the reason why the driver didn't get into upstream.

PR #312, opened 2026-06-22
DasharoPayloadPkg: Fix builds without TPM support

miczyg1:
The PR #282 broke builds where TPM_ENABLE is FALSE. The calls to the TPM2 command libraries querying supported PCR banks do not have a null library implementation, thus an attempt to build PlatformBootManagerLib without TPM fails due to the PCR bank checks. Compile out the PCR bank check if TPM is not enabled in the build.

### DIFF
--- a/DasharoPayloadPkg/DasharoPayloadPkg.dsc
+++ b/DasharoPayloadPkg/DasharoPayloadPkg.dsc
@@ -152,6 +152,9 @@
 
 [BuildOptions]
   *_*_*_CC_FLAGS                 = -D DISABLE_NEW_DEPRECATED_INTERFACES -Wno-stringop-overflow -mno-mmx -mno-sse
+!if $(TPM_ENABLE) == TRUE
+  *_*_*_CC_FLAGS                 = -D TPM_ENABLED
+!endif
 !if $(USE_CBMEM_FOR_CONSOLE) == FALSE
   GCC:RELEASE_*_*_CC_FLAGS       = -DMDEPKG_NDEBUG
   INTEL:RELEASE_*_*_CC_FLAGS     = /D MDEPKG_NDEBUG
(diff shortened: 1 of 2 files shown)


DasharoPayloadPkg/DasharoPayloadPkg.dsc
```diff
@@ -151,7 +151,11 @@
   DEFINE NETWORK_IPXE                   = FALSE
 
 [BuildOptions]
+!if $(TPM_ENABLE) == TRUE
+  *_*_*_CC_FLAGS                 = -D TPM_ENABLED -D DISABLE_NEW_DEPRECATED_INTERFACES -Wno-stringop-overflow -mno-mmx -mno-sse
```
SergiiDmytruk: Looked up the syntax for `[BuildOptions]` in [the docs](https://tianocore-docs.github.io/edk2-DscSpecification/draft/3_edk_ii_dsc_file_format/36_[buildoptions]_sections.html#36-buildoptions-sections) as I had questions about the semantics in other cases:

> Two types of assignment operators are permitted, the single equal "=" sign is used to append the string to the existing definition, while the double equal "==" sign is used to override any previous definition, replacing the content with just the string.

> Entries in [BuildOptions] sections are cumulative (as opposed to last line takes precedence in other section of the DSC file) provided the "==" character sequence is not used in statement later in the section. 

So no need for duplication, should be able to just append `-D TPM_ENABLED`.

miczyg1: https://github.com/Dasharo/edk2/pull/312/commits/4d7a970ca96f1803d34f228daaafa8a41e5e328f

Still works.

SergiiDmytruk (approved):

